PR #18, opened 2016-08-11
Number of CPUs should be a runtime argument, not an input.

tmooney:
The runtime will inform the tool how many cores it has to work with.


jasonwalker80:
+1
My experience with CWL is limited. It would appear that CPUs is now an argument, not an input.

PR #22, opened 2016-09-07
A CWL workflow for running Varscan Somatic detection.

tmooney:
Includes:
- samtools mpileup (with output marked "streamable")
- varscan somatic
- varscan processSomatic


jasonwalker80:
+2

PR #25, opened 2016-10-05
Combined variant detection workflow

tmooney:


jasonwalker80:
+2, I'll make a separate issue for adding dbSNP and COSMIC inputs to Mutect2

PR #35, opened 2016-11-18
CWL for running VEP

susannasiebert:


detect_variants/vep.cwl
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Ensembl Variant Effect Predictor"
+baseCommand: ["perl", "/usr/bin/variant_effect_predictor.pl"]
```
tmooney: Would it be worthwhile to be explicit that this is `/usr/bin/perl` (both here and in the two images [here](https://github.com/genome/docker-vep/blob/2f2499473e0eb5300a3f3397df3132990237211d/Dockerfile#L31) and [here](https://github.com/genome/docker-vep-cwl/blob/cf2298d86216af305f4be90b49b519f8ae099721/Dockerfile#L4))?

I ask since it just came up about the `PATH` being overridden by LSF's own ideas of what the environment should be.

susannasiebert: I think you are right. I will update accordingly.

PR #52, opened 2016-12-14
Limit variant calls from each detector to an ROI.

### LINKED ISSUE #42: Add optional target regions input
This should limit all variant callers to only calling variants within the target regions.

tmooney:
Closes #42.

This requires genome/docker-varscan-cwl#3 for region-limiting Varscan.


susannasiebert:
+1

dufeiyu:
Can mutect/split_interval_list.cwl be placed under detect_variants since it seems generic and reusable ?

tmooney:
I've pushed a commit to implement @dufeiyu's suggestion.

PR #53, opened 2016-12-15
Initial somatic pipeline.

tmooney:
Combines detect_variants with alignment (assuming #40, #51, and #52).

PR #55, opened 2016-12-15
Convert pipeline to use Strelka2.

### LINKED ISSUE #54: update Strelka to Strelka2
https://github.com/Illumina/strelka/releases/latest

tmooney:
Not quite tested, since my machine doesn't seem to have enough RAM to satisfy Strelka's detection.

- [x] Depends on genome/docker-strelka#4 and genome/docker-strelka-cwl#2.
- [x] Probably conflicts with #52.
- [x] Definitely conflicts with #53.

Closes #54.

### DIFF
--- a/detect_variants/detect_variants.cwl
+++ b/detect_variants/detect_variants.cwl
@@ -17,14 +17,14 @@ inputs:
         secondaryFiles: .bai
     interval_list:
         type: File
-    strelka_config:
-        type: File
     dbsnp_vcf:
         type: File?
         secondaryFiles: .tbi
     cosmic_vcf:
         type: File?
         secondaryFiles: .tbi
+    strelka_exome_mode:
+        type: boolean
 outputs:
     final_vcf:
         type: File
@@ -48,8 +48,8 @@ steps:
             reference: reference
             tumor_bam: tumor_bam
             normal_bam: normal_bam
-            config: strelka_config
             interval_list: interval_list
+            exome_mode: strelka_exome_mode
         out:
             [merged_vcf]
     varscan:
--- a/strelka/strelka.cwl
+++ b/strelka/strelka.cwl
@@ -2,35 +2,43 @@
 
 cwlVersion: v1.0
 class: CommandLineTool
-label: "strelka 1.0.15"
-baseCommand: "/usr/bin/cwl_helper.sh"
+label: "strelka 2.7.1"
+baseCommand: "/usr/bin/docker_helper.pl"
 requirements:
     DockerRequirement:
-        dockerPull: "mgibio/strelka-cwl:1.0.15"
+        dockerPull: "mgibio/strelka-cwl:2.7.1"
     ResourceRequirement:
         coresMin: 8
 arguments:
-    - { valueFrom: $(runtime.cores), position: 5 }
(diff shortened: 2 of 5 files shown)


strelka/strelka.cwl
```diff
@@ -2,35 +2,43 @@
...
 cwlVersion: v1.0
 class: CommandLineTool
-label: "strelka 1.0.15"
-baseCommand: "/usr/bin/cwl_helper.sh"
+label: "strelka 2.7.1"
+baseCommand: "/usr/bin/docker_helper.pl"
 requirements:
     DockerRequirement:
-        dockerPull: "mgibio/strelka-cwl:1.0.15"
+        dockerImageId: "mgibio/strelka-cwl:2.7.1"
```
susannasiebert: What does the change from dockerPull to dockerImageId accomplish?

susannasiebert:
+1 Looks fine as far as I can tell.

strelka/strelka.cwl
tmooney: I changed as part of testing with a locally tagged version (see also the `cwltool` `--disable-pull` option).  I should put it back 😄

detect_variants/detect_variants.cwl
```diff
@@ -17,14 +17,14 @@ inputs:
...
         type: File[]
-    strelka_config:
-        type: File
     dbsnp_vcf:
         type: File?
         secondaryFiles: .tbi
     cosmic_vcf:
         type: File?
         secondaryFiles: .tbi
+    strelka_exome_mode:
```
jasonwalker80: Are boolean required args or can this be optional with a default of false? Maybe I'm missing the cwl syntax. At the detect_variants.cwl level should we have a less specific `--exome` boolean arg instead of it being strelka specific? Currently strelka is the only caller that has different usage/inputs, but I wouldn't be surprised if others, ex. pindel, also have different workflows/usage based on this flag.

strelka/strelka.cwl
```diff
@@ -2,35 +2,43 @@
...
         type: File
         inputBinding:
-            position: 3
+            prefix: '--referenceFasta='
+            separate: false
+            position: 5
         secondaryFiles: .fai
-    config:
-        type: File
+    exome_mode:
```
jasonwalker80: Similar comment as above, is this a required or optional arg?

detect_variants/detect_variants.cwl
tmooney: This is currently required just as the config file was previously. I believe any input can be made optional by appending `?` to the type.

I would worry about giving people the false impression that setting this to `true` might automatically adjust cut-offs or tweak parameters for other tools.  I could possibly see hooking it up to a second tool's `--exome` flag if one happens to have one... but we might want to keep them separate even then.

strelka/strelka.cwl
tmooney: This is currently a required argument that must be included in the inputs.  In the `.yml`, it would appear as:
```yaml
exome_mode: true
```
or
```yaml
exome_mode: false
```

detect_variants/detect_variants.cwl
jasonwalker80: agree, this is fine as-is. No need to plan for future tools with exome modes until we need to support them.

PR #56, opened 2016-12-18
Add pindel to detect_variants

dufeiyu:
Implement pindel process related cwl to detect_variants cwl


jasonwalker80:
I setup pindel on Friday using the IDT exome target regions and it's still running this morning, >48 hours later. Run times on other callers:
Varscan - 1 hour
Strelka - 1.5 hour
Mutect (50 chunks) - Longest running chunk 9 hours with a median of ~3 hours.

We will need to implement a similar scatter/gather as Mutect or find another way to reduce run time.

jasonwalker80:
Looking at the pindel log, it appears it's still on chr1....
/gscuser/jwalker/git/HCC1395/arvados/logs/exome_pindel.out

pindel/pindel.cwl
```diff
@@ -0,0 +1,45 @@
...
+label: "pindel v0.2.5b8"
+baseCommand: ["/usr/bin/pindel", "-i", "pindel.config"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "mgibio/pindel-cwl:v0.2.5b8"
+    - class: InitialWorkDirRequirement
+      listing:
+          - entryname: 'pindel.config'
+            entry: |
+                $(inputs.normal_bam.path) 400    normal
```
jasonwalker80: For the sample names, we may want to use all caps `TUMOR` and `NORMAL` to match the other callers.

For the insert size, should this be an optional argument with a default of 400?  For example, the HCC1395 exome data has a median closer to 200bp.

pindel/pindel.cwl
```diff
@@ -0,0 +1,45 @@
...
+        secondaryFiles: .bai
+    normal_bam:
+        type: File     
+        secondaryFiles: .bai
+    reference:
+        type: File
+        inputBinding:
+            prefix: "-f"
+            position: 1
+        secondaryFiles: [".fai", "^.dict"]
```
jasonwalker80: Is the picard sequence dictionary used by pindel?

pindel/somaticfilter.cwl
```diff
@@ -0,0 +1,33 @@
...
+                hom=6
+                pindel2vcf=/usr/bin/pindel2vcf
+                reference=$(inputs.reference.path)
+                referencename=GRCh38DH
+                referencedate=20161216
+                output=$(runtime.outdir)/pindel.out.vcf
+inputs:
+    reference:
+        type: File
+        secondaryFiles: [".fai", "^.dict"]
```
jasonwalker80: Is the picard sequence dictionary used?

pindel/workflow.cwl
```diff
@@ -0,0 +1,67 @@
...
+cwlVersion: v1.0
+class: Workflow
+label: "pindel workflow"
+requirements:
+    - class: MultipleInputFeatureRequirement
+    - class: SubworkflowFeatureRequirement
+inputs:
+    reference:
+        type: File
+        secondaryFiles: [".fai", "^.dict"]
```
jasonwalker80: Same comment about picard sequence dictionary, is it used by the pindel workflow?

pindel/pindel.cwl
dufeiyu: I will change to NORMAL TUMOR. 400 is the hardcoded value in GMS pindel run. It can be optional if it varies.

pindel/pindel.cwl
dufeiyu: Nope. pindel doesn't use .dict. I will remove this copy/paste error.

pindel/somaticfilter.cwl
dufeiyu: No. I will remove it.

pindel/workflow.cwl
dufeiyu: No. I will remove it.

pindel/pindel_cat.cwl
```diff
@@ -0,0 +1,48 @@
...
+        type: File
+    insert_size:
+        type: int
+        default: 400
+outputs:
+    per_interval_pindel_out:
+        type: File
+        outputSource: cat/pindel_out
+steps:
+    sed:
```
jasonwalker80: I'd suggest an awk instead of sed.  The interval start coordinate needs to be converted to zero-based instead of one-based coordinates to make a real BED file.

detect_variants/detect_variants.cwl
```diff
@@ -51,13 +51,21 @@ steps:
...
+            normal_bam: normal_bam
+            interval_list: interval_list
     combine:
         run: combine_variants.cwl
         in:
             reference: reference
             mutect_vcf: mutect/merged_vcf
             strelka_vcf: strelka/merged_vcf
             varscan_vcf: varscan/merged_vcf
+            pindel_vcf: pindel/merged_vcf
```
tmooney: [combine_variants.cwl](https://github.com/dufeiyu/arvados_trial/blob/e6ac2280f36f255486e6534571791a74096050e8/detect_variants/combine_variants.cwl) will need to be updated to account for this change.

jasonwalker80:
Pindel running with 50 chunked interval lists, `-T 4` and `-w 10` took ~10 hours per chunk with ~15-25% CPU utilization with 4 CPUs and 4GB of RAM. I don't think `-w` is really effective when running with small exome ROI.  It might be best to run in parallel by chromosome and limit variants after.

dufeiyu:
This PR becomes messy and outdated. I just created a new PR https://github.com/genome/arvados_trial/pull/62 to replace this.  Close this PR.

PR #57, opened 2016-12-20
Workflow for exome QC

susannasiebert:
I should that I did not expose any of the optional inputs to these tools at this point. I will wait to do so until we know what the desired parameters are.

### DIFF
--- a/qc/collect_alignment_summary_metrics.cwl
+++ b/qc/collect_alignment_summary_metrics.cwl
@@ -1,0 +1,25 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "collect alignment summary metrics"
+baseCommand: ["/usr/bin/java", "-Xmx16g", "-jar", "/usr/picard/picard.jar", "CollectAlignmentSummaryMetrics"]
+arguments:
+    ["OUTPUT=", { valueFrom: $(runtime.outdir)/AlignmentSummaryMetrics.txt }]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "registry.gsc.wustl.edu/genome/picard-2.4.1-r:2"
+inputs:
+    bam:
+        type: File
+        inputBinding:
+            prefix: "INPUT="
+    reference:
+        type: File
+        inputBinding:
+            prefix: "REFERENCE_SEQUENCE="
+outputs:
+    alignment_summary_metrics:
+        type: File
+        outputBinding:
+            glob: "AlignmentSummaryMetrics.txt"
--- a/qc/collect_alignment_summary_metrics.cwl
+++ b/qc/collect_alignment_summary_metrics.cwl
@@ -14,10 +14,12 @@ inputs:
         type: File
         inputBinding:
             prefix: "INPUT="
+        secondaryFiles: [.bai]
     reference:
         type: File
         inputBinding:
             prefix: "REFERENCE_SEQUENCE="
+        secondaryFiles: [.fai]
 outputs:
     alignment_summary_metrics:
         type: File
--- a/qc/collect_hs_metrics.cwl
+++ b/qc/collect_hs_metrics.cwl
@@ -1,0 +1,34 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "collect HS metrics"
+baseCommand: ["/usr/bin/java", "-Xmx16g", "-jar", "/usr/picard/picard.jar", "CollectHsMetrics"]
+arguments:
+    ["O=", { valueFrom: $(runtime.outdir)/HsMetrics.txt }]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "registry.gsc.wustl.edu/genome/picard-2.4.1-r:2"
+inputs:
+    bam:
+        type: File
+        inputBinding:
+            prefix: "I="
+    reference:
+        type: File
+        inputBinding:
+            prefix: "R="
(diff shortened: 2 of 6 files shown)


qc/workflow_exome.cwl
```diff
@@ -0,0 +1,77 @@
...
+
+cwlVersion: v1.0
+class: Workflow
+label: "Exome QC workflow"
+requirements:
+    - class: SubworkflowFeatureRequirement
+inputs:
+    bam:
+        type: File
+        secondaryFiles: [^.bai]
```
tmooney: [This alignment output](https://github.com/susannasiebert/arvados_trial/blob/1c90d18ef4d5c6bb06c5b2db45d997cfb1a3e9dc/unaligned_bam_to_bqsr/workflow.cwl#L32) shows the bam index as being `.bai` rather than `^.bai`.

qc/collect_alignment_summary_metrics.cwl
```diff
@@ -0,0 +1,25 @@
...
+      dockerPull: "registry.gsc.wustl.edu/genome/picard-2.4.1-r:2"
+inputs:
+    bam:
+        type: File
+        inputBinding:
+            prefix: "INPUT="
+    reference:
+        type: File
+        inputBinding:
+            prefix: "REFERENCE_SEQUENCE="
```
tmooney: Would it be useful to have the indexes as secondary files in these inputs?  I haven't been able to track down a page of Picard docs that says much of anything about BAM indexes.

qc/workflow_exome.cwl
susannasiebert: Yeah, it's weird. When I look at an actual output directory of this workflow the file is named Final.bai. Not sure what is happening there.

qc/collect_alignment_summary_metrics.cwl
jasonwalker80: I'm not certain whether any of the indices are required inputs.  However, I do think the indices are used by some of the tools.  For instance, the .bai must be used to jump to regions of interest in the BAM when running CollectHsMetrics.  Also, the .fai is likely used when running CollectGcBiasMetrics.

qc/collect_hs_metrics.cwl
```diff
@@ -0,0 +1,35 @@
...
+        secondaryFiles: [.fai]
+    bait_intervals:
+        type: File
+        inputBinding:
+            prefix: "BAIT_INTERVALS="
+    target_intervals:
+        type: File
+        inputBinding:
+            prefix: "TARGET_INTERVALS="
+outputs:
```
jasonwalker80: Should we add the optional output for 'PER_TARGET_COVERAGE'? Is it possible with CWL to have optional outputs and how are those options triggered?

qc/workflow_exome.cwl
```diff
@@ -0,0 +1,85 @@
...
+        type: File
+        secondaryFiles: [.bai]
+    reference:
+        type: File
+        secondaryFiles: [.fai]
+    bait_intervals:
+        type: File
+    target_intervals:
+        type: File
+    roi_intervals:
```
jasonwalker80: I can't recall if we need a separate set of ROI intervals or if we should simply use the target_intervals.  I'm inclined to start with only target_intervals unless there is a need for a third set of intervals (outside of target/bait).

qc/collect_hs_metrics.cwl
tmooney: CWL has optional outputs using `?` in the type just like optional inputs.  They have [an example here](https://github.com/common-workflow-language/cwltool/blob/717b8384ea812870289e30f022bf85f5c1c7dcdb/cwltool/schemas/v1.0/v1.0/optional-output.cwl#L22).

jasonwalker80: Sorry @susannasiebert, I missed the comment above about not implementing optional features yet(which is fine). I was aware CWL had the ability to have optional outputs but I wasn't sure how they are connected to an input param to turn that feature on/off. Does the name of the file need to be provided as an input or a Boolean arg?

susannasiebert: I think either way is possible. The boolean arg option would be more user friendly IMHO. If the option is picked we can set PER_TARGET_COVERAGE to a hardcoded file.

qc/workflow_exome.cwl
susannasiebert: I'm totally fine with changing this so that select_variants/verify_bam_id runs on the target_intervals instead.

qc/collect_hs_metrics.cwl
jasonwalker80: sounds good. no edits requested for this since we can add optional inputs later.

qc/workflow_exome.cwl
jasonwalker80: Do we need a separate roi_intervals input? I'd suggest using the target_intervals for verifyBamID.

susannasiebert: If we use target_interval for verifyBamID than we don't need to roi_intervals. I will delete them.

dufeiyu:
@jasonwalker80  It seems picard CollectTargetedPcrMetrics generates some exome target metrics. Can we add it ?

jasonwalker80:
Do we need to run that in addition to CollectHsMetrics? From what I see it seems like CollectHsMetrics will collect all the QC metrics we need.  Is there something specific we are missing with Hs that we get with TargetedPcr?

dufeiyu:
You are correct. CollectHsMetrics does have what we need. I was not familiar with CollectHsMetrics and spoke too soon.

PR #62, opened 2017-01-03
Add pindel workflow to detect_variants

dufeiyu:
This PR implements pindel workflow parallelized by chromosome to detect_variants workflow. Pindel runs reasonably fast by parallelizing by chromosome. The largest chromosome 1 (249 mb) can finish within 1.5 hrs.

### DIFF
--- a/detect_variants/detect_variants.cwl
+++ b/detect_variants/detect_variants.cwl
@@ -25,6 +25,13 @@ inputs:
         secondaryFiles: .tbi
     strelka_exome_mode:
         type: boolean
+    pindel_chromosome_list:
+        type:
+            type: array
+            items: string
+    pindel_insert_size:
+        type: int
+        default: 400
 outputs:
     final_vcf:
         type: File
@@ -61,6 +68,17 @@ steps:
             interval_list: interval_list
         out:
             [merged_vcf]
+    pindel:
+        run: ../pindel/workflow.cwl
+        in:
+            reference: reference
+            tumor_bam: tumor_bam
+            normal_bam: normal_bam
+            interval_list: interval_list
+            chromosome_list: pindel_chromosome_list
+            insert_size: pindel_insert_size
+        out:
+            [merged_vcf]
     combine:
         run: combine_variants.cwl
         in:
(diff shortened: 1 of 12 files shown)


pindel/pindel.cwl
```diff
@@ -0,0 +1,63 @@
...
+      ramMin: 16000
+    - class: InitialWorkDirRequirement
+      listing:
+          - entryname: 'pindel.config'
+            entry: |
+                $(inputs.normal_bam.path) $(inputs.insert_size)    NORMAL
+                $(inputs.tumor_bam.path)  $(inputs.insert_size)    TUMOR
+arguments:
+    ["-w", "20",
+     "-T", "4",
```
jasonwalker80: Is there a CWL resource requirement that this will need 4 CPUs?

detect_variants/detect_variants.cwl
```diff
@@ -11,25 +11,32 @@ inputs:
...
         type: File?
-        secondaryFiles: .tbi
+        secondaryFiles: [".tbi"]
     cosmic_vcf:
         type: File?
-        secondaryFiles: .tbi
+        secondaryFiles: [".tbi"]
     strelka_exome_mode:
         type: boolean
+    pindel_chromosome_list:
```
jasonwalker80: Instead of adding a new input to determine which chromosomes we want to run pindel on, could we pull the chromosomes from the interval list?  For exome, we only run on targeted chromosomes.  For WGS, we will be providing an interval list for Mutect to limit variant detection to the autosomal+sex chromosomes of the genome.

pindel/pindel.cwl
dufeiyu: I will dig around.

detect_variants/detect_variants.cwl
dufeiyu: Good idea since interval_list seems required. I will make the change.

dufeiyu:
@jasonwalker80 change was made, tested and pushed. Can you review again ? thanks

detect_variants/detect_variants.cwl
```diff
@@ -11,32 +11,28 @@ inputs:
         secondaryFiles: [".fai", "^.dict"]
     tumor_bam:
         type: File
-        secondaryFiles: [".bai"]
+        secondaryFiles: [.bai]
```
jasonwalker80: You will likely have merge conflicts here.  Since GATK doesn't produce a .bam.bai file extension you will want to use `[^.bai]`

PR #63, opened 2017-01-12
CWL to run bam-readcount on a VCF

susannasiebert:
The helper.pl script in this CWL is mostly taken from https://github.com/ernfrid/fpfilter-tool/blob/master/fpfilter.pl. I was able to successfully run this CWL on some HCC1395 data but the resulting output files where empty. I might've not used the correct combination of files but I'm not sure.

This CWL will need to be wrapped/combined with a different step to run these on the different BAMs and output an file list yml for pVAC-Seq.

### DIFF
--- a/pvacseq/bam-readcount.cwl
+++ b/pvacseq/bam-readcount.cwl
@@ -1,0 +1,171 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "run bam-readcount"
+
+baseCommand: ["/usr/bin/perl", "helper.pl"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "mgibio/bam-readcount-cwl:0.7.4"
+    - class: InitialWorkDirRequirement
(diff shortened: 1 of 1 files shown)


pvacseq/bam-readcount.cwl
```diff
@@ -0,0 +1,174 @@
...
+            my $bam_file = $ARGV[2];
+            my $ref_fasta = $ARGV[3];
+            my $snp_output = $ARGV[4];
+            my $indel_output = $ARGV[5];
+
+            my $input = IO::File->new($file);
+            my @vcf_header = parse_vcf_header($input);
+            my $header_line = $vcf_header[-1];
+            chomp $header_line;
+            my @header_fields = split \"\\t\", $header_line;
```
jasonwalker80: I'm assuming the escapes are necessary as this is parsed as a string by bash? Otherwise, this would simply be `split "\t", $header_line;` right?

susannasiebert: That is correct.

PR #79, opened 2017-02-13
Switch expected Bam Index Name for Pindel Workflow

tmooney:
A few more places need `"^.bai"` to go along with the alternate naming standard.


dufeiyu:
+1

PR #82, opened 2017-02-14
Run CWL validation with Travis CI.

tmooney:
This should run `cwltool --validate` on each CWL file on master and in PRs.


susannasiebert:
+1

PR #84, opened 2017-02-15
Strelka2 Output 

jasonwalker80:
Update the paths to the strelka2 output files.


susannasiebert:
+1. I assume we don't need the index files from Strelka for the process_vcf.cwl step?

PR #85, opened 2017-02-15
Exome Alignment

jasonwalker80:
basic exome alignment and qc pipeline.

### DIFF
--- a/exome_alignment.cwl
+++ b/exome_alignment.cwl
@@ -32,7 +32,7 @@ inputs:
 outputs:
     bam:
         type: File
-        outputSource: alignment/bqsr_bam
+        outputSource: alignment/final_bam
     insert_size_metrics:
         type: File
         outputSource: qc/insert_size_metrics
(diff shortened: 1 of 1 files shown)


exome_alignment.cwl
```diff
@@ -0,0 +1,76 @@
...
+    alignment:
+        run: unaligned_bam_to_bqsr/workflow.cwl
+        in:
+            reference: reference
+            bams: bams
+            readgroups: readgroups
+            mills: mills
+            known_indels: known_indels
+            dbsnp: dbsnp
+        out: [bqsr_bam]
```
tmooney: The output of `unaligned_bam_to_bqsr/workflow.cwl` is named `final_bam`.

(It's a bummer that `cwltool --validate` doesn't check this--for subworkflows it only looks that the named `.cwl` file exists!)

jasonwalker80: Thanks for catching this. It should be addressed with the latest commit.

PR #92, opened 2017-02-21
Remove mutect index.

jasonwalker80:
Remove index after scattered mutect.


tmooney:
I think we need to add the index as a secondary file [here in mutect.cwl](https://github.com/jasonwalker80/arvados_trial/blob/4f13345d6ce6493a70cf3e3103aa78b130948fc3/mutect/mutect.cwl#L54).

PR #95, opened 2017-03-03
Remove Stdin from intervals_to_bed

jasonwalker80:
Remove stdin requirement and use grep/awk instead of perl.


susannasiebert:
+1

varscan/intervals_to_bed.cwl
```diff
@@ -1,30 +1,21 @@
...
     - class: DockerRequirement
       dockerPull: "ubuntu:xenial"
     - class: InitialWorkDirRequirement
       listing:
-          - entryname: 'helper.pl'
+          - entryname: 'helper.sh'
             entry: |
-                use feature qw(say);
-
-                for my $line (<STDIN>) {
```
tmooney: It might've been nice to make the change from Perl to `grep` + `awk` in a separate commit; if we'd wanted to "fix" the Perl script, changing this `<STDIN>` to `<>` would do it.

jasonwalker80:
See pull request #96 for the perl version.

PR #96, opened 2017-03-06
Remove stdin from intervals_to_bed - perl

jasonwalker80:
Use the perl version of the intervals_to_bed.cwl but remove the stdin requirement. Also add test data as a separate commit.


susannasiebert:
+1

Issue #38, opened 2016-12-05
Attempt running detect_variants.cwl with Toil


jasonwalker80:
As a test of the CWL portability and an evaluation of the Toil workflow system, attempt running our detect_variants.cwl workflow ideally using the LSF `--batchSystem` for job execution.

jasonwalker80:
From @tmooney 
> `cwltoil --help` includes this option:
> >>>--batchSystem BATCHSYSTEM
>                        The type of batch system to run the job(s) with,
>                        currently can be one of singleMachine, parasol,
>                        gridEngine, lsf or mesos'. default=singleMachine
> 
> [1:59]
> I did `pip install toil[cwl]` to get that command

tmooney:
`toil` appears to not support CWL v1.0.
```
$ cwltoil --batchSystem lsf workflow.cwl workflow-inputs.yaml 
...
Exception: Unrecognized version v1.0
```

I suppose that's what BD2KGenomics/toil#1340 is meant to address.

mr-c:
@tmooney You can get nearly complete CWL v1.0 support if you install the development release of toil

```
pip install --pre toil[cwl]
```

tmooney:
Cool, thanks!  The development release does seem to work better!  I ran into:
```
ERROR root: Unsupported requirement InitialWorkDirRequirement
```
for the [VCF indexing step](https://github.com/genome/arvados_trial/blob/0bae9acae07427fed1ee05a8dc74e071cbd548fb/detect_variants/index.cwl).  I see this feature was listed as known missing in BD2KGenomics/toil#1056.

mr-c:
@tmooney Glad to hear it. The lack of `InitialWorkDirRequirement` is possible to be worked around if absolutely necessary: use http://www.commonwl.org/v1.0/CommandLineTool.html#ShellCommandRequirement and symlink the input file into the working directly yourself.

Obviously it would be ideal if Toil supported `InitialWorkDirRequirement` natively :-)

jasonwalker80:
The toil_compatibility branch works with Toil at this time. It's currently under development as we add some new features, but this issue can be closed.

PR #102, opened 2017-03-22
Initial changes for running in toil.

tmooney:
* No docker images (for now).
* Stop using unsupported CWL features.
* Include some small example/test data.
* ...and more!


detect_variants/detect_variants.cwl
```diff
@@ -25,6 +25,10 @@ inputs:
         secondaryFiles: [.tbi]
     strelka_exome_mode:
         type: boolean
+    mutect_scatter_count:
+        type: int?
+    mutect_artifact_detection_mode:
```
tmooney: Yeah... I only added it out of suspicion that the optional parameters weren't working properly.  Now that more things are straightened out I could try making them optional again.

jasonwalker80:
Closes, #38

PR #107, opened 2017-03-22
Add VEP to toil detect_variants.cwl

susannasiebert:


susannasiebert:
I was planning on updating the example data detect variants yaml as part of this PR but decided against it for now since the current version only uses generic file names, and not actual existing file paths. I can add the vep_cache_dir input to it but the file wouldn't work in its current version since the paths don't point to anything. 

Are there other yaml files that need to be updated?

jasonwalker80:
This would be the main one I was thinking of:
https://github.com/genome/arvados_trial/blob/toil_compatibility/example_data/detect_variants/detect_variants.yaml

susannasiebert:
Ah yeah, and I was confused how this is working since it only uses file names, and not full file paths. Either way, I can add the vep_cache_dir I used to it as part of this PR.

example_data/detect_variants/detect_variants.yaml
```diff
@@ -19,3 +19,9 @@ tumor_bam:
   class: File
   path: H_NJ-HCC1395-HCC1395_exome.bam
 strelka_exome_mode: true
+vep_cache_dir:
+  class: File
+  path: /gscmnt/gc2764/cad/ssiebert/toil_test/inputs/vep_cache_location.txt
+synonyms_file:
+  class: File
+  path: /gscmnt/gc2764/cad/ssiebert/toil_test/inputs/chromAlias.ensembl.txt
```
jasonwalker80: There is a file in this dir `chromAlias.ensembl.txt` that only contains chr22. I'd point to that one.

susannasiebert: Thanks for the pointer. I updated the yaml. I also copied my vep_cache_location.txt into this location and amended the path accordingly. The txt still points to the actual path in my location but I don't think there is a good way around it since we don't want to make the cache files part of this repo.

PR #111, opened 2017-03-27
Add docm to detect_variants workflow

dufeiyu:
This PR is to add docm to detect_variant workflow.

### DIFF
--- a/detect_variants/combine_docm.cwl
+++ b/detect_variants/combine_docm.cwl
@@ -1,0 +1,35 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "CombineVariants (GATK 3.6)"
+baseCommand: ["/usr/bin/java", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "CombineVariants"]
+arguments:
+    ["-genotypeMergeOptions", "PRIORITIZE",
+     "--rod_priority_list", "filtered,docm",
+     "-o", { valueFrom: $(runtime.outdir)/combine_docm.vcf }]
+inputs:
+    reference:
+        type: File
+        inputBinding:
+            prefix: "-R"
+            position: 1
+        secondaryFiles: [.fai, ^.dict]
+    filtered_vcf:
+        type: File
+        inputBinding:
+            prefix: "--variant:filtered"
+            position: 2
+        secondaryFiles: [.tbi]
+    docm_vcf:
+        type: File
+        inputBinding:
+            prefix: "--variant:docm"
+            position: 3
+        secondaryFiles: [.tbi]
+outputs:
+    combine_docm_vcf:
+        type: File
+        outputBinding:
+            glob: "combine_docm.vcf"
+
--- a/detect_variants/combine_docm.cwl
+++ b/detect_variants/combine_docm.cwl
@@ -3,7 +3,11 @@
 cwlVersion: v1.0
 class: CommandLineTool
 label: "CombineVariants (GATK 3.6)"
-baseCommand: ["/usr/bin/java", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "CombineVariants"]
+baseCommand: ["/usr/bin/java", "-Xmx8g", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "CombineVariants"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 8000
+
 arguments:
     ["-genotypeMergeOptions", "PRIORITIZE",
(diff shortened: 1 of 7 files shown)


docm/GATK_haplotype_caller.cwl
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "HaplotypeCaller (GATK 3.6)"
+baseCommand: ["/usr/bin/java", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "HaplotypeCaller"]
```
jasonwalker80: I would include the java heap size in the command-line, ex. `"-Xmx8g", "-jar"`

docm/GATK_haplotype_caller.cwl
```diff
@@ -0,0 +1,48 @@
...
+    ["-gt_mode", "GENOTYPE_GIVEN_ALLELES",
+    "-o", { valueFrom: $(runtime.outdir)/docm_out.vcf }]
+inputs:
+    reference:
+        type: File
+        inputBinding:
+            prefix: "-R"
+            position: 1
+        secondaryFiles: [.fai, ^.dict]
+    normal_bam:
```
jasonwalker80: Do we want to genotype the normal or only the tumor to identify the presence of DoCM mutations? I was under the impression we only wanted to identify somatic events, in this case DoCM mutations unique to the tumor BAM/CRAM.

detect_variants/combine_docm.cwl
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "CombineVariants (GATK 3.6)"
+baseCommand: ["/usr/bin/java", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "CombineVariants"]
```
jasonwalker80: Same comment as below, add the heap size.

detect_variants/combine_docm.cwl
```diff
@@ -0,0 +1,35 @@
...
+arguments:
```
jasonwalker80: I would define at minimum a 4GB RAM resource requirement. On full exome data sets, this often failed without a minimum RAM requirement. It should be equal to or greater than the heap size above.

docm/GATK_haplotype_caller.cwl
dufeiyu: Dave is ok with this. We are looking for >5 reads and >1% VAF on tumor with normal as reference (not strict somatic).

docm/GATK_haplotype_caller.cwl
dufeiyu: I will make the change.

detect_variants/combine_docm.cwl
dufeiyu: I will make the change.

detect_variants/combine_docm.cwl
dufeiyu: I will add 8GB here.

docm/GATK_haplotype_caller.cwl
jasonwalker80: what do you mean "with normal as reference"? It seems unnecessary to run HaplotypeCaller on the normal if only the tumor genotypes are what we want. HaplotypeCaller doesn't have a somatic mode.

dufeiyu: According to Dave, he also checks the var count and vaf of normal if the tumor gets >5 reads and >1% VAF.

jasonwalker80: 👍

dufeiyu:
combine_docm.cwl is based on combine_variants.cwl. So I also added ram requirement and heap size for combine_variants.cwl.

dufeiyu:
There is a duplicated RAM requirement in select_variants.cwl. I removed it.

detect_variants/detect_variants.cwl
```diff
@@ -102,10 +105,39 @@ steps:
             vcf: combine/combined_vcf
         out:
             [filtered_vcf]
+    bgzip:
+        run: bgzip.cwl
+        in:
+            file: filter/filtered_vcf
+        out:
+            [bgzipped_file]
+    index:
```
tmooney: There are already steps in `detect_variants.cwl` named `bgzip` and `index`.  They'll need to be unique or else references like `bgzip/bgzipped_file` become ambiguous.

dufeiyu: good point. make the change.

PR #123, opened 2017-03-30
VEP changes

susannasiebert:


susannasiebert:
Nope, exactly the other way around. `local_cache: true` means that you have a cache file on your local disk somewhere.

susannasiebert:
I'm open to naming this something different if it's confusing.

example_data/detect_variants/vep_without_local_cache.yaml
```diff
@@ -0,0 +1,10 @@
+vcf:
+    class: File
+    path: exome.merged.fpfilter.pass.vcf.gz
+cache_dir:
```
jasonwalker80: If `local_cache: false` then do we need the VEP cache location?

susannasiebert: Yes, for the plugins. The Wildtype plugin is not available through VEP.

jasonwalker80: Ah, that makes sense.

jasonwalker80:
I think I understand the usage of local_cache now.  If we intend to run this with `local_cache: false` then I'd suggest providing the plugins in the example_data and adding a `vep_plugins_location.txt` that can be used in combination with `local_cache: false`.

susannasiebert:
Good idea about the example data. I will add that.

I'm not 100% happy with the naming of these inputs. VEP expects a directory that contains a `homo_sapiens` and a `Plugins` folder. The first will contain the reference cache(s) and the latter contains the Plugins. The parent directory is what we use as the `cache_dir` input. For `local_cache: true` we will still need a `cache_dir` with a `Plugins` folder underneath that containing the needed plugins. We might want to consider renaming the `cache_dir` to something more generic that encompasses in meaning both the reference caches and the plugins. I have no good idea what that could be. Also, would it be helpful to call the option `local_ref_cache` instead of `local_cache`?

jasonwalker80:
Should we put the plugins somewhere else? Perhaps the plugins need to go in the docker container itself rather than storing them in the cache dir.  Relevant line from Ensembl docs, "by default VEP includes the '~/.vep/Plugins' directory in the path".  Should we put plugins in ~/.vep/Plugins in the docker container? Then the directory is truly only the cache and the existing names are suitable.  cache_dir would then become optional and you could remove the local_cache option entirely.

susannasiebert:
That might work. I would need to do some testing.

susannasiebert:
@jasonwalker80 This is ready for merging. I tested just the vep step as well as the detect_variants.cwl successfully. Feel free to merge whenever.

detect_variants/detect_variants.cwl
```diff
@@ -36,9 +36,11 @@ inputs:
          type: File
          secondaryFiles: [.tbi]
     vep_cache_dir:
-        type: File
+        type: string?
```
jasonwalker80: Nice, so it validates as a string, but it just so happens to be a Directory(which is not supported).  Works for me!

susannasiebert: The only downside is that it won't check that the directory exists.

detect_variants/vep.cwl
```diff
@@ -6,20 +6,18 @@ label: "Ensembl Variant Effect Predictor"
 baseCommand: ["/usr/bin/perl", "/usr/bin/variant_effect_predictor.pl"]
```
susannasiebert: Added. I reran the vep.cwl and the output VCF now contains the DownstreamProtein and WildtypeProtein fields.

PR #124, opened 2017-03-30
VCF to TSV

jasonwalker80:
Initial version of a VCF to TSV converter.

### DIFF
--- a/detect_variants/variants_to_table.cwl
+++ b/detect_variants/variants_to_table.cwl
@@ -1,0 +1,46 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "SelectVariants (GATK 3.6)"
+baseCommand: ["/usr/bin/java", "-Xmx4g", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "VariantsToTable"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 6000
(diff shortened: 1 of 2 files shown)


detect_variants/variants_to_table.cwl
```diff
@@ -0,0 +1,46 @@
...
+            prefix: "-R"
+            position: 1
+        secondaryFiles: [".fai", "^.dict"]
+    vcf:
+        type: File
+        inputBinding:
+            prefix: "--variant"
+            position: 2
+        secondaryFiles: [.tbi]
+    fields:
```
jasonwalker80: Is there a concept of reserved words in CWL? I've seen fields used in examples where there are multiple optional inputs.

detect_variants/variants_to_table.cwl
```diff
@@ -0,0 +1,46 @@
...
+      type:
+        type: array
+        items: string
```
tmooney: I like the `type: string[]` syntax more, but either one should work.

jasonwalker80: I agree but I wasn't sure if that would work since I have these nested `type:` and `inputBinding:` to keep the array of args grouped on the command line.

detect_variants/variants_to_table.cwl
jasonwalker80: This worked. The input named `fields` didn't cause any problems.

PR #137, opened 2017-04-07
bgzip combine variants

jasonwalker80:
bgzip the combine variants output.


jasonwalker80:
Merging early since this is required for `toil_compatibility` to work.

PR #139, opened 2017-04-07
bgzip output for combine variants.

jasonwalker80:


jasonwalker80:
Merged early, third times a charm.

Issue #32, opened 2016-11-02
Create a Panel of Normals (PON) to be used with Mutect2


jasonwalker80:
Dave Spencer will provide a list of 10 or so samples for an initial implementation of PON.  List will be provided via @dufeiyu

jasonwalker80:
This is now waiting on an MGI workorder to generate GRCh38DH alignments:
https://imp-lims.gsc.wustl.edu/entity/setup-work-order/2853720

jasonwalker80:
https://jira.gsc.wustl.edu/browse/ITSD-919

jasonwalker80:
I believe we can still take advantage of the `scatter: interval_list` but will need to run GATK with the normal CRAM inputs as the tumor inputs to GATK. I'm not 100% sure if the sample names will need to be unique or if they can all be called `TUMOR`. After we have mutect run (with `artifact_detection_mode: true` for each input CRAM/Sample we will combine the VCFs. The final output is a merged VCF that will be the PoN input to downstream workflows, ie. detect_variants

jasonwalker80:
The attempt with 5 samples has succeeded. It appears when running Mutect2 with artifact detection mode, that it outputs the real sample name instead of tumor. This prevents a collision of `TUMOR` samples in the VCF.

Location of example PON VCF:
`/gscmnt/gc2764/cad/tmooney/toil_test/out/pon/panel_of_normals.vcf.gz`

The other 5 should be available to make a PON from all 10 samples.

jasonwalker80:
I made a small chr22 region VCF for testing:
/gscmnt/gc2764/cad/HCC1395/toil/refseq/GRCh38DH_chr22/panel_of_normals.vcf.gz

jasonwalker80:
@dufeiyu created a new PoNs with 10 samples.

PR #140, opened 2017-04-10
Updated Panel-of-Normals workflow for CRAM-based mutect

tmooney:
This is part of #32.


panel_of_normals/combine_variants.cwl
```diff
@@ -2,14 +2,16 @@
...
 requirements:
-    - class: DockerRequirement
-      dockerPull: "dbmi/gatk-docker:v1" #GATK 3.6 at a specific container revision
+    - class: ResourceRequirement
+      ramMin: 8000
+      tmpdirMin: 25000
 arguments:
-    ["-setKey", "null",
-     "--filteredAreUncalled", 
+    ["-minN", "2",
```
jasonwalker80: I think it's fine for now to use a minimum number of samples equal to 2. Depending on the number of input VCFs, this may need to change in the future.

PR #142, opened 2017-04-11
Add optional Panel-of-Normals VCF input to DetectVariants.

tmooney:
Output of my test run with a PON input specified is currently in `/gscmnt/gc2764/cad/tmooney/toil_test/out/dv/`.


dufeiyu:
+1

Issue #128, opened 2017-04-04
Normalize Variants


jasonwalker80:
Left Align indels
multiallelic sites
MNPs

jasonwalker80:
See this bam-readcount issue concerning normalization of indels:
https://github.com/genome/bam-readcount/issues/40

In that issue they list another tool for normalizing variants (including decomposition of biallelic block substitutions):
http://genome.sph.umich.edu/wiki/Vt

jasonwalker80:
Continue with GATK LeftAlignAndTrimIndels for version 1.

jasonwalker80:
We have been running with LeftAlign for many months. However, there are issues with A and R INFO tags in the VCF when using the --splitMultiAllelic option in GATK. Instead, we will turn off that option and add the `vt decompose` step immediately after LeftAlign. 

Then let's close this old issue. Switching to `vt` for normalization would require a revalidation of the pipelines and a comparison for type1 and type2 errors.

susannasiebert:
I think we can close this issue, correct?

PR #143, opened 2017-04-11
Left-Align Indels before running the FP filter.

tmooney:
This is part of #128.


tmooney:
Using our small test set each detector ends up with `0 variants were aligned`, so it runs but isn't doing too much.

Issue #144, opened 2017-04-11
Build an RNA-Seq workflow


jasonwalker80:
This wiki outlines one workflow for using the new Tuxedo suite of tools to get differential expression:
https://github.com/genome/arvados_trial/wiki/RnaSeq

jasonwalker80:
The first iteration needs to run:
1) HiSAT2
2a) RNA-seq QC
2b) feature-count

That's it for now.

PR #151, opened 2017-04-13
Start of an RnaSeq workflow: Alignment

tmooney:
This represents step 1 of #144.

One quirk of this version of the implementation: the `reference_index` parameter should point to a file whose name matches the name of the index files but without the extensions (`.1.bt` - `.8.bt`).  This placeholder file can be empty or have whatever contents, but should be small since it will be copied by `toil` alongside the other index files.

(To run, this currently depends on genome/docker-rnaseq#1 to provide the necessary environment.)

### DIFF
--- a/rnaseq/bam_to_fastq.cwl
+++ b/rnaseq/bam_to_fastq.cwl
@@ -1,0 +1,28 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Picard: BAM to FASTQ"
+baseCommand: ["/usr/bin/java", "-Xmx4g", "-jar", "/opt/picard/picard.jar", "SamToFastq"]
+arguments: [ {valueFrom: "F=$(runtime.outdir)/read1.fastq"},
+             {valueFrom: "F2=$(runtime.outdir)/read2.fastq"} ]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 6000
+      tmpdirMin: 25000
+inputs:
+    bam:
+        type: File
+        inputBinding:
+            prefix: "I="
+            separate: false
+            position: 1
+outputs:
+    fastq1:
+        type: File
+        outputBinding:
+            glob: "read1.fastq"
+    fastq2:
+        type: File
+        outputBinding:
+            glob: "read2.fastq"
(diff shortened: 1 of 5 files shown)


rnaseq/bam_to_fastq.cwl
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env cwl-runner
```
jasonwalker80: Is this the same as https://github.com/genome/cancer-genomics-workflow/blob/master/unaligned_bam_to_bqsr/revert_to_fastq.cwl

rnaseq/workflow.cwl
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "RnaSeq alignment workflow"
+requirements:
+    - class: MultipleInputFeatureRequirement
+    - class: ScatterFeatureRequirement
+    - class: SubworkflowFeatureRequirement
+inputs:
```
jasonwalker80: Does HiSAT2 take additional meta-data like sample, library, flow cell, etc.

rnaseq/bam_to_fastq.cwl
tmooney: Yep! Looks pretty much the same, but the path to picard is different.

rnaseq/workflow.cwl
tmooney: There are two relevant options:

```
--rg-id <text>
Set the read group ID to <text>. This causes the SAM @RG header line to be printed, with <text> as the value associated with the ID: tag. It also causes the RG:Z: extra field to be attached to each SAM output record, with value set to <text>.

--rg <text>
Add <text> (usually of the form TAG:VAL, e.g. SM:Pool1) as a field on the @RG header line. Note: in order for the @RG line to appear, --rg-id must also be specified. This is because the ID tag is required by the SAM Spec. Specify --rg multiple times to set multiple fields. See the SAM Spec for details about what fields are legal.
```

tmooney:
Once #157 is merged, the picard bam_to_fastq steps should become pretty much the same, so we should be able to combine those.  I see #159 for the other issue.

PR #161, opened 2017-05-03
Add read-group tag support to the HISAT2 CWL.

### LINKED ISSUE #159: Add readgroup info to HiSAT2 rnaseq workflow

tmooney:
This uses `scatterMethod: dotproduct`, so each `instrument_data_bam` parameter must have a corresponding `read_group_id` and array of `read_group_fields` listed in the same order.

Closes #159.


tmooney:
Interesting thought!  We probably could make our own common metadata inputs and use either an adaptor script or some code with an `InlineJavascriptRequirement` to munge it into the tool's parameters as needed.

PR #163, opened 2017-05-03
Set mark duplicates memory request to 40 GB.

tmooney:
This step tells Java to use 16GB heap and sambamba to use 18GB, so make sure there's enough to handle both of those plus overhead.


malachig:
+1 I agree this is a good starting point.  Hopefully more than this is not needed.

PR #171, opened 2017-05-19
Revert "Add option for adding HGVS annotation to VEP"

susannasiebert:
Reverts genome/cancer-genomics-workflow#166

Issue #170, opened 2017-05-19
Fix --hgvs option with VEP


jasonwalker80:
Options:
a) [temporary] Rollback the commit that added hgvs
b) modify workflow to include the FASTA with existing API version
c) Update to v87 (which requires an updated cache)

In standalone testing I've been using Ensembl API v87 like so:
`/usr/bin/perl /gscmnt/gc2764/cad/HCC1395/arvados/software/ensembl-vep/vep.pl -i /gscmnt/gc2764/cad/HCC1395/arvados/exome.merged.fpfilter.pass.vcf.gz --offline --af_exac --coding_only --hgvs --cache --dir /gscmnt/gc2764/cad/HCC1395/arvados/software/VEP_cache --format vcf --vcf --plugin Downstream --plugin Wildtype --symbol --terms SO --flag_pick -o /gscmnt/gc2764/cad/HCC1395/arvados/exome.merged.fpfilter.pass.annotated.vcf.g`

jasonwalker80:
For option C, see #173

jasonwalker80:
Updated to v87

PR #172, opened 2017-05-23
Fix hgvs parameter

susannasiebert:
This supersedes PR #171. It implements option B of #170.

### DIFF
--- a/detect_variants/vep.cwl
+++ b/detect_variants/vep.cwl
@@ -25,6 +25,8 @@ inputs:
         inputBinding:
             prefix: "-i"
             position: 1
+    reference:
+        type: File
     cache_dir:
         type: string?
         inputBinding:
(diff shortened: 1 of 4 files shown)


detect_variants/vep.cwl
```diff
@@ -25,6 +25,8 @@ inputs:
         inputBinding:
             prefix: "-i"
             position: 1
+    reference:
+        type: File
```
tmooney: Is the reference only required when `--hgvs` is specified?  If so, should this be optional (or possibly a [dependent parameter](http://www.commonwl.org/v1.0/UserGuide.html#Record_inputs,_dependent_and_mutually_exclusive_parameters))?

jasonwalker80: Should the reference also be optional? Is it only needed when the `--hgvs` option is set to true in this API version (86)?

susannasiebert: I made the reference optional. However, with that change if someone sets `hgvs` to `true` and doesn't specify a reference, the workflow will still dispatch but VEP will fail. 

I tried to implement Tom's suggestion of dependent_parameters but it seems like you can't make the dependent parameter set optional. I tried a few things with Tom's help but couldn't get it to work.

susannasiebert:
@tmooney I like this suggestion. @jasonwalker80 what do you think?

susannasiebert:
@tmooney The way we can get around that, I think is to keep the outside hgvs parameter and if it is set to true, set the hgvs_reference for the vep step to the main reference.

PR #176, opened 2017-06-13
Add several parameter inputs to varscan and mutect cwl

dufeiyu:
So we can change parameters via input yaml files. This PR relies on https://github.com/genome/docker-cle/pull/39.

### DIFF
--- a/detect_variants/detect_variants.cwl
+++ b/detect_variants/detect_variants.cwl
@@ -32,6 +32,24 @@ inputs:
         type: int?
     mutect_artifact_detection_mode:
         type: boolean?
+    mutect_max_alt_allele_in_normal_fraction:
+        type: float?
+    mutect_max_alt_alleles_in_normal_count:
+        type: int?
+    varscan_strand_filter:
+        type: int?
+        default: 0
+    varscan_min_coverage:
+        type: int?
+        default: 8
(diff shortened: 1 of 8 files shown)


detect_variants/detect_variants.cwl
```diff
@@ -32,6 +32,24 @@ inputs:
         type: int?
     mutect_artifact_detection_mode:
         type: boolean?
+    mutect_max_alt_allele_in_normal_fraction:
+        type: float?
+    mutect_max_alt_alleles_in_normal_count:
+        type: int?
+    varscan_strand_filter:
+        type: int?
+        default: 0
```
tmooney: Does this workflow still work if we only store the defaults in `varscan/varscan_somatic.cwl` rather than have copies in each layer of the CWL?

dufeiyu: hmm, to my understanding, the upper layer CWLs: detect_variants.cwl, varscan/workflow.cwl, varscan.cwl should work without the defaults set since the bottom layer CWL varscan_somatic.cwl has the defaults set. These parameter inputs need to be optional across the layers. I will make the change and test it if this is correct.

dufeiyu:
@jasonwalker80 ok, I will update the detect_variants.yaml.

PR #179, opened 2017-06-21
Change reference input type to `string` instead of `File`

susannasiebert:
I'm still trying to get a successful run of the full detect variants workflow.


susannasiebert:
I was able to run the detect variants workflow successfully with these changes.

jasonwalker80:
+2

PR #181, opened 2017-06-29
Update vep.cwl to v87

susannasiebert:


susannasiebert:
I updated the .travis.yml to install the required dependency for cwltool. The travis tests are now passing.

PR #187, opened 2017-07-10
Add StringTie to RNAseq workflow

susannasiebert:


susannasiebert:
Not needed after all.

PR #190, opened 2017-07-13
Add concordance cwl for future integration

mnneveau:
add a concordance directory holding the concordance tool CWL for future integration into the overall workflow


jasonwalker80:
Please see https://github.com/genome/concordance

concordance/concordance.cwl
```diff
@@ -0,0 +1,34 @@
...
+            position: 4
+        secondaryFiles: [.fai, "^.dict"]
+    snp:
+        type: File
+        inputBinding:
+            position: 5
+    output_file_name:
+        type: string?
+        inputBinding:
+            position: 6
```
tmooney: Based on the description of https://github.com/genome/concordance, it looks like this should have a prefix in its inputBinding for the parameter name:
```
prefix: "--output_geno"
```

concordance/concordance.cwl
```diff
@@ -0,0 +1,34 @@
...
+        default: "concordance_report.txt" 
+outputs:
+    output_file:
+        type: File
+        outputBinding:
+            glob: "*.txt"
```
tmooney: It'd be good to add a final newline to this file (if only to stop Github from complaining :smile:)

concordance/concordance.cwl
mnneveau: @tmooney should be fixed now

concordance/concordance.cwl
```diff
@@ -0,0 +1,43 @@
...
+        inputBinding:
+            position: 7
+            prefix: "--output_geno" 
+outputs:
+    output_file:
+        type: File
+        outputBinding:
+           glob: "$(inputs.output_file_name)"
+    output_geno_file:
+        type: File[]
```
tmooney: If this output file is only optionally produced, it should be optional here, too, to avoid the workflow crashing when it isn't present.

mnneveau:
thanks - that should be fixed now too

PR #199, opened 2017-09-08
initial revision of a singleton RNA-seq workflow, aka. no merge.

jasonwalker80:


jasonwalker80:
@chrisamiller is handling this in a better way. this PR is closed without merge.

PR #206, opened 2017-10-11
Add cle_annotated_vcf_filter step and strelka_cpu_reserved parameter

dufeiyu:
This PR is needed for several changes mentioned in https://jira.gsc.wustl.edu/browse/ITDEV-5037 before IT can release the IDT exome assay. It relies on https://github.com/genome/docker-cle/pull/50
This change has been tested ok in wdl_toil hybrid workflow run. @jasonwalker80 can you review it ?


strelka/strelka.cwl
```diff
@@ -9,7 +9,7 @@ requirements:
         coresMin: 8
         ramMin: 4000
 arguments:
-    [ { valueFrom: $(runtime.cores), position: 1 },
+    [ { valueFrom: $(inputs.cpu_num), position: 1 },
```
jasonwalker80: As I understand the issue, the problem is with LSF scheduling the jobs with 8 cores requested. This change will only modify the number passed to the strelka command.  Instead of modifying the number of cpus strelka expects, don't you want to modify the number of CPUs requested? See line9 for the resource requirements used when scheduling the LSF job.  

Changing the number of CPUs that strelka thinks it has available will result in twice the run time as before.

detect_variants/detect_variants.cwl
```diff
@@ -63,6 +66,9 @@ inputs:
         type: boolean?
     hgvs_annotation:
         type: boolean?
+    filter_annotated_vcf:
```
jasonwalker80: The name of this workflow step is a bit ambiguous. What is being filtered? There are other steps that are hard or soft filtering.

detect_variants/detect_variants.cwl
```diff
@@ -27,6 +27,9 @@ inputs:
         secondaryFiles: [.tbi]
     strelka_exome_mode:
         type: boolean
+    strelka_cpu_num:
```
jasonwalker80: should this be something like `strelka_cpu_reserved` or `min_strelka_cpus`? Do you want to go ahead and parameterize the number of threads/CPU strelka *thinks* it's using in the command line?

detect_variants/annotated_vcf_filter.cwl
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "annotated_vcf_filter"
+arguments: [
+    "/usr/bin/perl", "/usr/bin/annotated_vcf_filter.pl",
```
jasonwalker80: call `docm_and_coding_indel_selection.pl`

detect_variants/detect_variants.cwl
jasonwalker80: how about `cle_indel_filter`

detect_variants/detect_variants.cwl
```diff
@@ -263,11 +274,30 @@ steps:
             vcf: bgzip/bgzipped_file
         out:
             [indexed_vcf]
+    annotated_vcf_filter:
```
jasonwalker80: `cle_filtered_vcf` this is or is not filtered depending on the `cle_filter` boolean flag.

detect_variants/detect_variants.cwl
```diff
@@ -27,6 +27,9 @@ inputs:
         secondaryFiles: [.tbi]
     strelka_exome_mode:
         type: boolean
+    strelka_cpu_num:
+        type: int?
+        default: 4
```
jasonwalker80: change default to 8 since this value is passed to the strelka runWorkflow.py

strelka/strelka.cwl
```diff
@@ -9,7 +9,7 @@ requirements:
         coresMin: 8
```
jasonwalker80: since lookup values and variables may not work in the resourceRequirements section we could hard-code this to 4.

dufeiyu:
@jasonwalker80 Changes were made per review. wdl-toil test on 10% downsampling HCC1395 is fine. Can you check again ?  Thanks.

PR #212, opened 2017-10-23
Add FP Filter to Varscan, add DoCM to tumor-only DV workflow.

tmooney:
This makes it more similar to the Somatic Varscan workflow.


detect_variants/tumor_only_detect_variants.cwl
```diff
@@ -104,7 +126,7 @@ steps:
     bam_readcount:
         run: ../pvacseq/bam_readcount.cwl
```
dufeiyu: bam_readcount.cwl output changed in https://github.com/genome/cancer-genomics-workflow/pull/223

PR #215, opened 2017-10-25
Run pindel header grep right after cat_all using stream pipe

dufeiyu:
During toil_cwl run, sometimes all_chromosome_pindel.out becomes binary file mysteriously, which makes grep failing and therefore empty pindel output. This fix will make pindel cwl more robust.


pindel/cat_all.cwl
```diff
@@ -2,13 +2,18 @@
 cwlVersion: v1.0
 class: CommandLineTool
 baseCommand: ['/bin/cat']
-stdout: "all_chromosome_pindel.out"
+requirements:
+    - class: ShellCommandRequirement
+arguments: [
+    { shellQuote: false, valueFrom: "|" },
+    "/bin/grep", "ChrID", "/dev/stdin"
```
tmooney: Is the `/dev/stdin` part of the magic of making this work?  Usually it's more common to use `-` to indicate to read from the standard input.

dufeiyu: hmm, I basically "copied" this ```/dev/stdin``` from your "rnaseq/hisat2_align.cwl" as example. This works in my testing toil run. Do I need change this to ```-```?

tmooney: I copied that from somewhere else 😄.  I think it probably works as-is, but it would also work as an `-` or with that parameter removed altogether.

dufeiyu: ok, I will just keep as-is since it works in the testing run. Thanks.

PR #216, opened 2017-10-25
Provide reference to Picard CollectInsertSizeMetrics

tmooney:
Picard was failing with an error about not being able to find `chr1` even though it's in the reference.  Some advice on the GATK forum was to explicitly pass the reference file when using CRAM, so this implements that suggestion.


jasonwalker80:
#222 had these changes. closing.

PR #221, opened 2017-10-26
Add a downsample workflow for exome qc.

jasonwalker80:


jasonwalker80:
This does not work with the current version of Picard 2.4.1. I get the following error with and without the R= option
```
Exception in thread "main" java.lang.IllegalStateException: A valid CRAM reference was not supplied and one cannot be acquired via the property settings reference_fasta or use_cram_ref_download
```
I'm going to attempt updating docker-cle to Picard 2.14.0 to see if it fixes the issue. This blog/FAQ suggests that full CRAM support in all Picard commands was still buggy around the time of the 2.4.1 release:
https://gatkforums.broadinstitute.org/gatk/discussion/8291/markduplicates-in-cram-files-a-valid-cram-reference-was-not-supplied#latest

jasonwalker80:
This should be replaced by #225. It runs the Downsample in parallel, uses BAM files which avoids updating Picard and you get the actual duplication rate after downsampling.

PR #223, opened 2017-10-31
Add step to run add_bam_readcount_to_vcf_helper.py

dufeiyu:
Add a step to run add_bam_readcount_to_vcf_helper.py so AD, AF, DP info from bam_readcount can be added to vcf format fields. A real toil test is running now.


dufeiyu:
The real toil test passes with the code change. Now ready to merge.

PR #227, opened 2017-11-03
Add add_bam_readcount_to_vcf to tumor_only workflow

susannasiebert:
I ran a test for the add_bam_readcount_to_vcf.cwl with the included add_bam_readcount_to_vcf.yaml. I can move the test data files to the example_data directory, if desired. I haven't tested the changes to the tumor only workflow.

### DIFF
--- a/detect_variants/add_bam_readcount_to_vcf.cwl
+++ b/detect_variants/add_bam_readcount_to_vcf.cwl
@@ -5,17 +5,26 @@ class: CommandLineTool
 label: "add bam_readcount info to vcf"
 
 baseCommand: ["/usr/bin/python3", "/usr/bin/add_bam_readcount_to_vcf_helper.py"]
-arguments: 
-    ["TUMOR,NORMAL", $(runtime.outdir)]
+arguments:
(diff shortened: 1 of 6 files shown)


susannasiebert:
I ran this locally using the `/gscmnt/gc2585/fulton/model_data/cc994b38f1454a009117e1751b7152f3/build2b978d7996cd4cf096381a1edc78884a/inputs.yaml` yaml and it finished successfully. Output is at `/gscmnt/gc2764/cad/ssiebert/toil_test/out/lsf4/`

jasonwalker80:
I see `AF` so that's good!

susannasiebert:
Yeah, the test workflow I did ended after add_bam_readcount_to_vcf. I'm running it locally all the way through just to make sure there aren't any downstream failures but I think this should be ok to merge.

susannasiebert:
I had to remove certain fields from the variants to table step but I got it to succeed. Successful output is at `/gscmnt/gc2764/cad/ssiebert/toil_test/out/lsf5`

@jasonwalker80 can you check that the reduced list of field is ok for this workflow.

PR #228, opened 2017-11-06
Add ExAC filter

susannasiebert:


detect_variants/exac_filter.cwl
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "ExAC filter"
+baseCommand: ["/usr/bin/perl", "/opt/vep/ensembl-vep/filter_vep.pl"]
+requirements:
+    - class: ShellCommandRequirement
```
tmooney: Would this work without the `ShellCommandRequirement`?

susannasiebert: That is a good question. I'm not sure.

PR #229, opened 2017-11-07
Updates to Single-Sample DV based on Initial Tests

tmooney:
This adds some additional filtering and updates a few defaults.

~This depends on #228.~ Rebased after merge of #228.

### DIFF
--- a/detect_variants/coding_variant_filter.cwl
+++ b/detect_variants/coding_variant_filter.cwl
@@ -1,0 +1,22 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Coding Variant filter"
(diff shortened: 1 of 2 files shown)


detect_variants/add_bam_readcount_to_vcf.cwl
```diff
@@ -18,7 +18,6 @@ inputs:
             items: File
         inputBinding:
             prefix: ""
-            itemSeparator: ","
```
susannasiebert: For tumor-normal multi-sample workflows we probably still need this, right? Does it cause an error in the tumor-only workflow to run it with the itemSeparator?

tmooney: I was kind of wondering about that... I can take this commit out.  If it really is causing issues we should find find a different solution that works for both.

PR #236, opened 2017-12-07
New tool to run the GATK HaplotypeCaller and wrapper scatter workflow.

tmooney:


jasonwalker80:
is there an example yml to show how this would be run?

tmooney:
I didn't include one, but something like this would do it (if, for example, you wanted to use `chr1`, `chr22`, and then `chrX` and `chrY` together in one file):
```yaml
reference: '/path/to/all_sequences.fa'
cram:                                                                                                                                                                                                                                                                 
    class: File                                                                                                                                                                                                                                                       
    path: '/path/to/my.cram'                                                                                                                                                           
emit_reference_confidence: 'GVCF'                                                                                                                                                                                                                                     
gvcf_gq_bands: ['5', '20', '60']                                                                                                                                                                                                                                             
intervals: [['chr1'],['chr22'],['chrX', 'chrY']]                                                                                                                                                                                             
dbsnp_vcf:                                                                                                                                                                                                                                                            
    class: File                                                                                                                                                                                                                                                       
    path: '/path/to/Homo_sapiens_assembly38.dbsnp138.vcf.gz'
```

PR #238, opened 2017-12-19
CWLs for pVACseq, pVACfuse, and pVACvector

susannasiebert:
A few caveats:
- the pVACseq CWL does not implement the additional_input_files parameter. We would need to parse this file and import the listed files as an InitialWorkDirRequirement. Instead I will work in implementing pVACseq using VCF genotype fields for coverage, vaf, and expression since that is what we want to do in the long run anyway
- toil/docker doesn't support X11 forwarding so we are not able to draw the vector visualization. I implemented a change in pVACvector to check whether the display is set and only then draw the visualization. This means that pVACvector runs done in docker will not create the visualization but at least they won't fail.


susannasiebert:
I also had to remove the iedb_install_directory option because toil doesn't support directories or InitialWorkDirRequirement so short of pulling in every single IEDB file separately, we can't make the IEDB install available. Ideally we would be able to distribute IEDB with our docker container. @malachig should we contact IEDB about this?

susannasiebert:
I was able to add the IEDB directory to a toil run, using the `Directory` data type. However, I'm encountering errors while running IEDB. I tried messing with it for half a day and am giving up for now.

susannasiebert:
The error I'm getting is:
```
File "/private/var/lib/cwl/stge10a4e74-a0c6-4a5c-a9f0-9a5e854195b7/iedb/mhc_i/src/../method/netmhc-4.0-executable/netmhc_4_0_executable/__init__.py", line 33, in predict
    process = Popen(cmd, stdout=PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
[job test.cwl] completed permanentFail
```
when IEDB is trying to sub out to the individual prediction algorithms.
I'm not exactly sure why this is happening. It could be a permissions error. It could be that IEDB isn't quite set up correctly since the install gets copied into the container instead of unpacking the installation tarball and running the configure script. 
I think I will merge in this PR as is and we can discuss further steps at a later time.

chrisamiller:
> Ideally we would be able to distribute IEDB with our docker container. @malachig should we contact IEDB about this?

Yeah someone would have to have a conversation with IEDB about distribution in Docker. As is, with the academic license, I suspect that it's a no-go.  That said, we should go ahead and build a docker container with IEDB and keep it in the private docker repo. Plenty of internal users that could benefit, and then the work is done if we can make it public.

PR #240, opened 2017-12-28
Update vep.cwl to use v90

susannasiebert:
This also brings the master version of vep.cwl in line with the toil_compatibility branch.

### DIFF
--- a/detect_variants/vep.cwl
+++ b/detect_variants/vep.cwl
@@ -5,7 +5,6 @@ class: CommandLineTool
 label: "Ensembl Variant Effect Predictor"
 baseCommand: ["/home/vep/src/ensembl-vep/vep", "-I", "/home/vep/Plugins"]
 requirements:
-    - class: ShellCommandRequirement
     - class: InlineJavascriptRequirement
     - class: DockerRequirement
       dockerPull: "mgibio/vep-cwl:v90"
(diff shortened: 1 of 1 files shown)


detect_variants/vep.cwl
```diff
@@ -3,22 +3,20 @@
 cwlVersion: v1.0
 class: CommandLineTool
 label: "Ensembl Variant Effect Predictor"
-baseCommand: ["/usr/bin/perl", "/usr/bin/variant_effect_predictor.pl"]
-
+baseCommand: ["/home/vep/src/ensembl-vep/vep", "-I", "/home/vep/Plugins"]
 requirements:
+    - class: ShellCommandRequirement
```
tmooney: (I see it's in the `toil_compatibility` version, too, but) what's using the `ShellCommandRequirement`?

susannasiebert: I'm always confused when to use which requirements. I think I might've added it for the `cache_dir` and `hgvs` `valueFrom` inputs. But I guess it's not needed for that?

tmooney: Yeah, the `valueFrom` expressions are the `InlineJavascriptRequirement`.  The most common reason we need `ShellCommandRequirement` is for pipes.

susannasiebert: Ok, I removed it.

PR #242, opened 2017-12-29
Toil compatibility

chrisamiller:
New workflow that includes transcript quantification with kallisto


susannasiebert:
I just realized that this PR fails integration tests. The tests run `find . -name '*.cwl' | xargs -n 1 cwltool --validate`. You'll want to go over the errors and see if you can resolve them.

chrisamiller:
my bad - just pushed a fix

chrisamiller:
There is a mismatch between the version of cwltool in the rnaseq docker container (/usr/local/bin/cwltool 1.0.20161221171240) and the one in the travis container that does validation (/home/travis/virtualenv/python2.7.14/bin/cwltool 1.0.20171227212058) (Ugh...).  This causes some things to validation in the rnaseq container that fail here under the newer validator. Looking into both updating it in the rnaseq container and fixing the validation error.

chrisamiller:
I pushed some changes that fix the validation error in Toil 3.12, and have also tested this with LSF. As soon as travis-ci approves, I think this is ready to go

rnaseq/bam_to_trimmed_fastq_and_hisat_alignments.cwl
```diff
@@ -2,58 +2,68 @@
...
+    min_readlength:
         type: int
+    read_group_id:
+        type: string
+    read_group_fields:
+        type:
+            type: array
+            items: string
+    reference_index:
+        type: File
```
tmooney: Was removing the `secondaryFiles` here intentional?  They're still present in `hisat2_align.cwl`.  (I guess they aren't required to be repeated if this was still working!)

rnaseq/kallisto.cwl
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Kallisto: Quant"
+baseCommand: ["/usr/bin/kallisto"] 
+requirements:
+    - class: ShellCommandRequirement
```
tmooney: Is this really using the `ShellCommandRequirement`? It should only be included if the command must be run through a shell instead of directly `exec`ed.

rnaseq/transcript_to_gene.cwl
```diff
@@ -0,0 +1,28 @@
...
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Kallisto: TranscriptToGene"
+baseCommand: ["/usr/local/bin/Rscript"]
+arguments: [
+    "/usr/src/transcript_to_gene.R"
+]
+requirements:
+    - class: ShellCommandRequirement
```
tmooney: Same question about `ShellCommandRequirement` here.  Generally best to avoid it unless we need pipes or other shenanigans!

rnaseq/hisat2_align.cwl
```diff
@@ -12,7 +12,6 @@ requirements:
 arguments: [
     "-p", $(runtime.cores),
     "--dta",
-    "--rna-strandness", "RF",
```
jasonwalker80: Do we want to remove this? I'm not sure it has much effect, but all of our standard production RNA-seq is TruSeq Stranded Total RNA.

rnaseq/workflow.cwl
```diff
@@ -1,75 +0,0 @@
...
-            trimming_min_readlength: trimming_min_readlength
-        out:
-            [aligned_bam]
-    merge:
-        run: merge.cwl
-        in:
-            bams: align/aligned_bam
-        out:
-            [merged_bam]
-    stringtie:
```
jasonwalker80: Are we no longer running StringTie on RNA-seq? Kallisto has similar results, but I presume some users are still interested in alignment based expression estimates. Did this move to another workflow?

rnaseq/kallisto.cwl
chrisamiller: it is not - removed.

rnaseq/transcript_to_gene.cwl
chrisamiller: removed

rnaseq/hisat2_align.cwl
chrisamiller: I'm running a second test pipeline now so I can compare stranded and unstranded params on the same data (in kallisto and hisat). Will let you know if it matters much or not.

rnaseq/workflow.cwl
chrisamiller: all of workflow.cwl has been removed and replaced by "workflow_firststrand_hisat_stringtie_kallisto.cwl".  We're already planning another cwl with STAR, so I want to make sure they're labelled appropriately until we have a better way to track these.

rnaseq/hisat2_align.cwl
chrisamiller: reposting from slack:
I just did a comparison of stranded vs unstranded params on a small test data set.  The alignments are pretty much identical with HISAT2, but about 0.3% of genes have substantially altered expression via kallisto

So yeah, small difference in general, but I'm still more comfortable having the correct flags set (with the understanding that a mixup is not the end of the world)

jasonwalker80: I agree, we could make it an optional parameter and then the higher level workflows would need to pass the appropriate value.

rnaseq/workflow.cwl
jasonwalker80: Sounds good. I don't see the larger workflow CWL in this PR. Am I missing it or will it be in a follow up PR?

chrisamiller: Pushing it in about 2 minutes - finishing the last tests now

chrisamiller:
Added optional boolean parameters 'firststrand' and 'secondstrand' that do the right thing in kallisto and hisat. Tested it using the new LSF params in toil, and I think it's all set.

PR #244, opened 2018-01-04
Point to specific versioned GATK JAR for HaplotypeCaller.

tmooney:
This is to get around a bug in 3.6.  See also genome/docker-cle#64.


tmooney:
(This should not be merged until genome/docker-cle#64 has been merged and a new image built on Dockerhub.)

PR #253, opened 2018-02-09
adding bam indexing

chrisamiller:


rnaseq/workflow.cwl
```diff
@@ -45,9 +45,10 @@ inputs:
     secondstrand:
        type: boolean?
 outputs:
-    merged_bam:
-      type: File
-      outputSource: merge/merged_bam
+    final_bam:
+        type: File
+        outputSource: index_bam/indexed_bam
+        secondaryFiles: [.bai, ^.bai]
```
tmooney: Do both of these `secondaryFiles` get produced? I only see one for the output of [the `index_bam.cwl`](https://github.com/chrisamiller/cancer-genomics-workflow/blob/2bb7ed2fbea9428cb0d6b71859dc1922cf1f8090/detect_variants/index_bam.cwl#L19).

PR #254, opened 2018-02-13
Add a little buffer space between system memory limit and sambamba param.

tmooney:
12G > 12000M, so this was unlikely to work out before.


chrisamiller:
+3

PR #257, opened 2018-02-20
gnomADe_AF Filter for tumor-only

jasonwalker80:
update to use gnomADe_AF as population allele frequency filter


detect_variants/gnomADe_AF_filter.cwl
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "MAX_AF filter"
```
tmooney: Might want to update this label :smile:

jasonwalker80:
Label updated. As for the defaults I'm starting to second guess my changes to `tumor_only_detect_variants.cwl`. The default behavior (with no `custom_gnomad_vcf` input is to still get the allele frequencies using the cache and hence MAX_AF and MAX_AF_POPS is still appropriate. In fact, it almost seems like we need two distinct workflows, 1) with a custom VCF 2) using the cache, each with their own defaults....

susannasiebert:
This looks good to me but I agree with what you are saying about the defaults.

jasonwalker80:
A little bit of a rewrite to use a "generic" filter CWL with an input string for the correct field in the VEP annotated VCF.

jasonwalker80:
@tmooney Did you use this PR as-is or did you use only part of the commits? I'm not sure if I should merge this after the changes to make the allele frequency an input parameter.

tmooney:
I used the `gnomADe_AF_filter.cwl` filter from this PR, but otherwise did not use this PR, as it adds a different filter to the workflow instead.  I added the gnomad filter in place of the AF filter for my runs.

jasonwalker80:
Let's go with #265 instead of the "generic" AF CWL

Issue #297, opened 2018-06-14
Optionally run coverage for a specific list of genes.


jasonwalker80:
For most applications this should be the ACMG gene list. We will need % at 0x, % at 10x, % at 20x, and then mean coverage for each exon of the genes.

jasonwalker80:
There is some discussion on exactly how to determine the genomic coordinates for the ACMG gene list.

One approach can be found here: https://jira.ris.wustl.edu/browse/BIO-2714

jasonwalker80:
Add Per-Target and Per-Base executions of CalculateHsMetrics with independent input interval lists for each. Ex. ACMG genes and ClinVar variants.

PR #298, opened 2018-06-18
Run CollectHsMetrics with additional coverage targets.

tmooney:
This change makes the per-base and per-target coverage required.

This is part of #297.

### DIFF
--- a/qc/workflow_exome.cwl
+++ b/qc/workflow_exome.cwl
@@ -82,7 +82,8 @@ steps:
         in:
             cram: cram
             reference: reference
-            metric_accumulation_level: picard_metric_accumulation_level
+            metric_accumulation_level:
+                valueFrom: "ALL_READS"
             bait_intervals: bait_intervals
             target_intervals: target_intervals
             per_target_coverage:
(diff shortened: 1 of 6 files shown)


qc/workflow_exome.cwl
```diff
@@ -75,10 +75,40 @@ steps:
...
+                default: false
+        out:
+            [hs_metrics]
+    collect_per_base_metrics:
+        run: collect_hs_metrics.cwl
+        in:
+            cram: cram
+            reference: reference
+            metric_accumulation_level: picard_metric_accumulation_level
+            bait_intervals: bait_intervals
```
jasonwalker80: I would suggest using the per_base_intervals for both the target_intervals and bait_intervals. Otherwise, the bait_intervals would be all over the genome/exome. The per-base intervals are likely individual sites that may or may not be covered by a bait. If we use the WES bait intervals, the Picard metrics for efficiency of design, etc. will also be skewed or not useful. The ultimate solution would be to intersect the baits with the per_base_interval input. However, I don't think it's necessary at this time.

qc/workflow_exome.cwl
```diff
@@ -75,10 +75,40 @@ steps:
...
+                default: true
+        out:
+            [per_base_coverage_metrics]
+    collect_per_target_metrics:
+        run: collect_hs_metrics.cwl
+        in:
+            cram: cram
+            reference: reference
+            metric_accumulation_level: picard_metric_accumulation_level
+            bait_intervals: bait_intervals
```
jasonwalker80: same comment as above, but use the per_target_intervals as bait_intervals input.

PR #312, opened 2018-07-05
New Germline WGS workflow for alignment, QC, HaplotypeCaller, VEP.

tmooney:
Based on the Exome workflow, but with the WGS alignment process instead!


tmooney:
As suggested, I'll revise this PR after #317 and #319 have been addressed.  That will simplify this new workflow quite a bit!

jasonwalker80:
FYI, #317 and #319 are both merged.

tmooney:
Updated this PR to account for the other merged changes.

dufeiyu:
+1

PR #313, opened 2018-07-10
Update add_bam_readcount_to_vcf.cwl to use vcf-readcount-annotator

susannasiebert:


detect_variants/add_bam_readcount_to_vcf.cwl
```diff
@@ -4,32 +4,28 @@ cwlVersion: v1.0
 class: CommandLineTool
 label: "add bam_readcount info to vcf"
 
-baseCommand: ["/usr/bin/python3", "/usr/bin/add_bam_readcount_to_vcf_helper.py"]
+baseCommand: ["vcf-readcount-annotator"]
+requirements:
+    - class: InlineJavascriptRequirement
```
tmooney: `InlineJavascriptRequirement` isn't necessary to use `$(runtime.outdir)` (as seen [here](https://github.com/genome/cancer-genomics-workflow/blob/00aa67b1fe0400cde023272dfa472bc56bae0314/qc/collect_insert_size_metrics.cwl#L8), for example).

susannasiebert: Argh, I always get that wrong. It's now removed.

susannasiebert:
Tests are now passing.

detect_variants/add_bam_readcount_to_vcf.cwl
```diff
@@ -4,32 +4,26 @@ cwlVersion: v1.0
 class: CommandLineTool
 label: "add bam_readcount info to vcf"
 
-baseCommand: ["/usr/bin/python3", "/usr/bin/add_bam_readcount_to_vcf_helper.py"]
+baseCommand: ["vcf-readcount-annotator"]
```
tmooney: We've often been specifying the full path to these rather than trusting they're in the `$PATH`.

susannasiebert: It's a pip package. Not sure where that would get installed to in our setup.

tmooney: We could try `which vcf-readcount-annotator` inside the container to find the full path.  The thought behind my comment is to make sure it works when using the scrubbed user environment, so if that's still functioning we needn't worry :smile:

susannasiebert: I ran the exome test as described here https://confluence.ris.wustl.edu/pages/viewpage.action?spaceKey=CI&title=Toil#Toil-PipelineTest. Not sure if that is sufficient.

tmooney: That ought to do fine!  Hopefully we'll eventually set up an automated test for that.

Issue #315, opened 2018-07-12
Ensure the Histogram PDF is saved for CollectInsertSizeMetrics


jasonwalker80:
https://github.com/genome/cancer-genomics-workflow/blob/e4c851d65f460e8f48ca184120044fce72cb2433/qc/collect_insert_size_metrics.cwl#L31

PR #317, opened 2018-07-12
Save the insert size histogram during QC.

tmooney:
This would close issue #315, except that the new workflow in #312 will also need to be updated, depending which PR gets merged first.


jasonwalker80:
I added an invite to @johnegarza as a collaborator for this repo. I'm hopeful I can then assign you this for review. It would be a nice exercise in code review to go over @tmooney's changes.

Issue #316, opened 2018-07-12
Add minimum quality parameters


jasonwalker80:
Add inputs for MINIMUM_MAPPING_QUALITY and MINIMUM_BASE_QUALITY. Other versions of these workflows are using values of 1 to include all reads during coverage assessment:
https://github.com/genome/cancer-genomics-workflow/blob/e4c851d65f460e8f48ca184120044fce72cb2433/qc/collect_hs_metrics.cwl#L13

PR #318, opened 2018-07-17
2 new input parameters in collect_hs_metrics

johnegarza:
Added in 2 new inputs to fix issue #316


tmooney:
We ran the tests, both with and without the new optional parameters.  Annotated TSV unsurprisingly came out identical.  Parameters definitely impacted the HS Metrics file!

Issue #214, opened 2017-10-25
Make -q and -b bam-readcount parameters user-configurable


susannasiebert:

chrisamiller:
Closed by #320

PR #320, opened 2018-07-18
Made -q and -b bam_readcount flags user definable

johnegarza:
Modified ./pvacseq/bam_readcount.cwl and all upstream dependencies to optionally accept -q (minimum mapping quality) and -b (minimum base quality) flags. Should close issue #214


johnegarza:
I wasn't sure about defining defaults- we commented on this in /genome/docker-cle#76. At the moment, there are default values to simplify logic in that PR's python script. One of these default values was already present in the script (-b=20) before my PR, and is different from the default value within the actual tool (-b=0), so I left it as 20 to avoid breaking/changing old workflows that used this.

Issue #202, opened 2017-09-14
Add --rf option to StringTie


jasonwalker80:
The rest of the workflow assumes firststrand RNA-seq library inputs. Newer versions of StringTie have added the option to specify the library strandedness. We should update the params passed to StringTie accordingle, ex. add --rf.

chrisamiller:
This has been resolved

PR #321, opened 2018-07-19
Stringtie now accepts strandedness flag

johnegarza:
Added in two optional parameters that can add --rf or --fr flags to stringtie; this should resolve issue #202. Possibly builds on https://github.com/genome/cancer-genomics-workflow/commit/3cd5ee0003faacec8a92a4b50faff89a2d8f5ac9

### DIFF
--- a/rnaseq/stringtie.cwl
+++ b/rnaseq/stringtie.cwl
@@ -14,20 +14,30 @@ arguments: [
     "-n", $(runtime.cores)
 ]
 inputs:
+    firststrand:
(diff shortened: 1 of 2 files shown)


chrisamiller:
Consider changing line 5 of workflow.cwl where it talks about first-strandedness.  Other than that, looks good 👍

johnegarza:
I think that should work; I followed the example of the commit I referenced in the top level comment and made both parameters optional.

PR #323, opened 2018-07-24
Removed unused parameter to fix travis validation warning

johnegarza:
Travis CI logs from the cwl validator tool showed 2 secondary outputs expected from a tool run from the rnaseq workflow cwl, but the tool itself only generates one output. Removed the unnecessary expected output from the workflow script.


johnegarza:
Ran through the workflow with test data as described in this [tutorial](https://confluence.ris.wustl.edu/display/CI/Toil) and verified that this update does not change the results.

PR #325, opened 2018-07-26
adding kallisto fusion preprocessing step

chrisamiller:
This step adds very little overhead, and provides the input needed to run pizzly for fusion detection.


chrisamiller:
I had the same thought, but was struggling to find another term. Lemme figure out something

chrisamiller:
renamed and tested - will merge tomorrow if there are no further objections

Issue #186, opened 2017-07-05
Add QC to RNA-seq workflow


jasonwalker80:
Run Picard CollectRnaSeqMetrics on the aligned CRAM file to generate basic RNA-seq QC metrics. This will require converting the GTF to a refFlat format and also providing an intervals format for Ribosomal and Mitochondrial annotation.

jasonwalker80:
Example: ```gmt gtf to-ref-flat --input-gtf-file=/gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/Homo_sapiens.GRCh38.90.gtf --output-file=/gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/Homo_sapiens.GRCh38.90.refFlat.txt```

```
grep rRNA /gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/Homo_sapiens.GRCh38.90.gtf > /gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/Homo_sapiens.GRCh38.90.rRNA.gtf

gmt gtf to-intervals --gtf-file=/gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/Homo_sapiens.GRCh38.90.rRNA.gtf --interval-file=/gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/Homo_sapiens.GRCh38.90.rRNA.interval_list --seqdict-file=/gscmnt/gc2560/core/model_data/2887491634/build21f22873ebe0486c8e6f69c15435aa96/seqdict/seqdict.sam

```

jasonwalker80:
Using `mgibio/cle` docker image:
```
[...]

PR #326, opened 2018-07-26
rnaseq workflow now contains a tool to generate qc metrics

johnegarza:


rnaseq/workflow.cwl
```diff
@@ -36,14 +36,27 @@ inputs:
...
+        type: File
+    ribosomal_intervals:
+        type: File
+    strand:
+        type: string
+    metrics_output_file:
+        type: string
+    chart_output_file:
+        type: string
+    bam:
```
chrisamiller: bam shouldn't be an input here, right? QC steps should be taking the aligned_bam produced by the hisat step

chrisamiller:
What are the refflat and ribosomal intervals inputs?  Can those be derived from the existing annotations (ref fasta? GTF?) quickly in real time? If not, we need to update our annotation directories with those files and make it clear how to produce them.   Just trying to make sure we keep the overhead of adding new annotations/genome builds as low as possible.

tmooney:
Examples of how to create the necessary input files are in #186.  It seems possible we could convert from existing inputs within the workflow, but in the absence of shortcutting we've tended to avoid that--especially in cases where it depends on `gmt` tools.

One middle-ground option might be to have the `gmt`s run in `process_inputs.pl` like the BED-to-interval-list lookups happen now for ROIs, but I think it's preferable to generate them once for the annotation set and have them available.  Once upon a time our annotation importers would take care of something like this--we really should keep track of what derivative files we need to have in place somehow!

johnegarza:
As @tmooney said, it's possible to generate some of those files as the workflow runs, but we're not entirely sure of the best way to do that yet. @jasonwalker80 suggested hardcoding pre-generated files for now to keep the workflow functional, and opening an issue (#327) to remind us to take care of it later.

chrisamiller:
Makes sense to me.  Can I request that in the short term, you drop the needed files into these directories so that they're easy to find?

/gscmnt/gc2560/core/GRC-human-build38_human_90_38_U2AF1_fix/rna_seq_annotation/
/gscmnt/gc2560/core/GRC-mouse-build38_mouse_92_38/rna_seq_annotation/
(looks like the human one may already be there, but the mouse one is definitely missing!)

If we have those two pipelines up and running, that covers at least 95% of what we'll be processing!

chrisamiller:
+1

johnegarza:
I'll get to that as soon as I'm able to- at the moment I don't have the permissions needed to access that part of the filesystem or run gmt.

rnaseq/workflow.cwl
```diff
@@ -36,14 +36,20 @@ inputs:
...
        type: File
     firststrand:
        type: boolean?
     secondstrand:
        type: boolean?
+    refFlat:
+        type: File
+    ribosomal_intervals:
+        type: File
+    strand:
```
tmooney: Could we calculate this value from the existing `firststrand` and `secondstrand` inputs?

jasonwalker80: If possible, calculating this property would avoid much of the confusion associated with strand. For example, if it's the firststrand setting in StringTie, then you'd want SECOND_READ_TRANSCRIPTION_STRAND for this tool. 

Here is a summary table from the RNA-seq PLoS paper with the Griffith Lab:
https://github.com/griffithlab/rnaseq_tutorial/blob/e6677df4f76c7c023c725de6202cb9e421aa459a/manuscript/supplementary_tables/supplementary_table_5.md

jasonwalker80:
All remaining tasks are logged in #333 and #327

PR #332, opened 2018-09-10
adding filtering to detect variants

chrisamiller:
This turned out way bigger than I'd like, but the essential changes are detailed below:

1) add a filter_vcf step to the detect_variants workflow.  This takes in a annotated vcf, and runs any number of filtering steps, then returns a filtered VCF (with failing variants labeled in the FILTER field.) Some steps may add additional annotation where needed

2) pulled existing filtering steps (CLE filter, gnomAD filter) into the above step

3) Added a mapq0 filter to remove regions with at least 15% reads with mapping quality zero (by default)

4) did a ton of cleanup to make tests run for the somatic_exome_workflow.cwl, which calls all of the above.  This leverages the existing HCC1395 bams in the example_data directory, and adds some interval lists and such that are needed for the full test that goes from bam -> fastq -> aligned data -> detect_variants  all the way to a final filtered annotated VCF.   It still fails at present because the tumor and normal cram files stomp on each other with the same file name


exome_alignment.cwl
```diff
@@ -17,7 +17,7 @@ inputs:
     known_indels:
         type: File
         secondaryFiles: [.tbi]
-    dbsnp:
+    dbsnp_vcf:
```
tmooney: If we're renaming this one, should we also be changing `known_indels` and `mills`? (`omni_vcf` already has "vcf" in the name.)  I suppose a general question: Is it helpful including filetypes in input names?  We could rename `reference` to `reference_fasta` while we're at it!  (I think I'd argue against doing that... but what does that mean for the others?)

chrisamiller: I agree that we should look for consistency here! I did some quick greps and must have missed some.  

My general rule has been - if I'm putting together a yaml file and it's not clear what file is needed, then more specificity is needed.  There are a variety of different dbSNP formats, but a reference is almost certainly a fasta (that said, I am generally in favor of being more descriptive, not less, so reference_fasta would be fine!).  

Can we go ahead and punt additional renaming to a separate PR, since this one is bloated enough?

tmooney: Yeah--we can certainly worry about renaming other inputs later!  I'll gladly +1 after my other comment is dealt with (which should ensure Travis CI is satisfied) 😄

chrisamiller:
I believe all the renaming is done now and the tests pass

jasonwalker80:
+1, thanks Chris!

PR #337, opened 2018-09-11
Filter_vcf_coding_variant to definitions/tools

apaul7:


apaul7:
@tmooney also fixed tool call I missed here

PR #340, opened 2018-09-11
moved filter_vcf_allele_freq to new definitions/tools directory

apaul7:
could not locate a path which needed to be updated after the move for this tool


apaul7:
@susannasiebert added the docker requirement

PR #341, opened 2018-09-11
Move location of bam_to_cram.cwl and index_cram.cwl

susannasiebert:
As part of this I also updated the bisulfite workfow.cwl to use the bam_to_cram.cwl and index_cram.cwl instead of the reimplementation that calls out to a helper script. @chrisamiller, as a result the helper script https://github.com/genome/docker-bisulfite/blob/master/bam_to_cram can be deleted from the bisulfite docker image.


chrisamiller:
No objections here.  I assume that these sections are the parts that run subworkflows in different docker containers? 
```
    - class: DockerRequirement
         dockerPull: "mgibio/cle"```

susannasiebert:
@chrisamiller yes

PR #357, opened 2018-09-11
 Rename bisulfite/merge.cwl to merge_bams.cwl and move to definitions/tools

susannasiebert:
This also deleted the merge.cwl's in the rnaseq and unaligned_bam_to_bqsr folders since they were doing the same thing


susannasiebert:
@tmooney good catch, that totally fell through the cracks. It's now updated.

PR #368, opened 2018-09-11
Move vcf2bed.cwl to definitions/tools

susannasiebert:


susannasiebert:
@tmooney I didn't rename this one because no other vcf2bed tool exists. For pileup I wanted to make sure it is differentiated from mpileup which is why I changed the name to biscuit_pileup.

PR #369, opened 2018-09-11
Moved revert_to_fastq to tools, added docker

johnegarza:


susannasiebert:
Good catch @tmooney. I didn't realize that. I would remove it then.

johnegarza:
Sounds good to me, I'll delete and merge it in

PR #375, opened 2018-09-11
Move trim_fastq cwls to definitions/tools

susannasiebert:
The output of the trim_fastq cwls for bisulfite and rnaseq differ in that one is a list and the other splits out fastq1 and fastq2. This separation is necessary because of subtle changes in the outer workflows. Therefore, both were ported to definitions/tools with slightly different names.


susannasiebert:
@tmooney I consolidated the two trim_fastq tools.

PR #396, opened 2018-09-11
Move combine_variant to def/tools, add docker

dufeiyu:
Note there is a same named one under panel_of_normals that uses different parameters


dufeiyu:
@susannasiebert panel_of_normal is likely a stand-alone, not automated pipeline. I lean toward to rename the one under panel_of_normals to be pon_combine_variants.cwl and keep this one the name as is.

PR #403, opened 2018-09-12
 Rename GATK_haplotype_caller.cwl and germline_gatk_haplotypecaller.cwl and move to definitions/tools

susannasiebert:


tmooney:
I bet we could consolidate these two tools! 
Possibly something like this would preserve the existing behavior:
```yaml
cram:
  ...
  position: 3
normal_cram:
  type: File?
  ...
    position: 2
```
(Dealing with the third GATK HaplotypeCaller tool (currently `definitions/tools/gatk_haplotypecaller.cwl`) might take more work, though.)

susannasiebert:
I consolidated the two docm ones under docm_gatk_haplotype_caller.cwl. There are too many differences with the gatk_haplotype_caller.cwl so I will make an issue to consolidate those.

PR #405, opened 2018-09-12
moved strelka/add_gt.cwl to tools, added dockerRequirement

apaul7:


apaul7:
@tmooney I updated the name as you suggested

PR #426, opened 2018-09-12
Migrate varscan

johnegarza:


definitions/subworkflows/varscan.cwl
```diff
@@ -82,14 +82,14 @@ steps:
         out:
             [snvs, indels]
     process_somatic_snvs:
-        run: ../definitions/tools/varscan_processsomatic.cwl
```
susannasiebert: I wish we had renamed this workflow to `varscan_process_somatic.cwl`. I must've missed that in the original tools PR. Now might be a good time to do that?

johnegarza: I'll fix that now

PR #429, opened 2018-09-12
Update strelka subworkflows and workflows

susannasiebert:


susannasiebert:
@tmooney ah yes, I could update the path here, but I also already have the fp_filter PR open

PR #435, opened 2018-09-13
Umi/molecular_alignment_workflow to subworkflows

apaul7:


apaul7:
Figured it would be easiest to move `molecular_qc_workflow.cwl` at the same time as the other ones instead of waiting for this to merge

PR #440, opened 2018-09-13
Docm/workflow.cwl to subworkflows with rename

apaul7:
`docm/workflow.cwl` to `definitions/subworkflows/docm_cle.cwl`


dufeiyu:
Just talk to Alex. The def/tools/somatic_docm_filter.cwl is better named as def/tools/docm_cle_filter.cwl and def/subworkflows/docm_somatic.cwl is renamed to def/subworkflows/docm_cle.cwl

PR #442, opened 2018-09-13
Add pvacseq pipeline

susannasiebert:
This is uses the outputs from detect_variants and rnaseq to prep the vcf and run pvacseq. 

Some paths will need to be updated as we continue to migrate the cwls.

We also need to decide which options to expose and which ones to hardcode.

pvacseq.cwl currently uses Zach's internal docker image on the registry which contains the upcoming 1.1 version of pvactools (staging branch) and includes IEDB. Once we make an official pvactools release we need to make an "official" image on the registry. We can't put it on docker hub (yet) since we don't have permissions to redistribute IEDB.

These cwl validates but I haven't run it yet.


definitions/tools/pvacseq.cwl
```diff
@@ -0,0 +1,174 @@
...
+        outputBinding:
+            glob: "combined/*.combined.parsed.tsv"
+    combined_filtered_epitopes:
+        type: File?
+        outputBinding:
+            glob: "combined/*.filtered.tsv"
+    combined_ranked_epitopes:
+        type: File?
+        outputBinding:
+            glob: "combined/*.filtered.condensed.ranked.tsv"
```
tmooney: Just a warning, currently optional file plus wildcard glob yields an error in Cromwell, as seen here: broadinstitute/cromwell#4004  One workaround is presented there. If you can compute the filename a solution like #329 also works.

definitions/tools/vcf_readcount_annotator.cwl
```diff
@@ -7,7 +7,7 @@ label: "add bam_readcount info to vcf"
 baseCommand: ["vcf-readcount-annotator"]
 requirements:
     - class: DockerRequirement
-      dockerPull: mgibio/cle
+      dockerPull: mgibio/vcf_annotation_tools-cwl
```
tmooney: If we're switching to smaller containers can we start to version this?

susannasiebert: Done.

PR #446, opened 2018-09-13
Migrate qc exome

johnegarza:


johnegarza:
Let's merge #435 in first, and then I'll resolve the conflicts here

PR #447, opened 2018-09-14
Detect variants pipelines

johnegarza:


johnegarza:
Merging now to finish up the migration, but I'll open up an issue to take care of the suggested renaming as soon as that's complete.

PR #449, opened 2018-09-14
Umi pipelines

johnegarza:


dufeiyu:
As what cwl labels suggest, these two can be better renamed as umi_duplex_alignment.cwl and umi_molecular_alignment.cwl

PR #454, opened 2018-09-14
Mv germline subworkflow

johnegarza:


dufeiyu:
This should be def/subworkflows/varscan_germline_workflow.cwl to differentiate def/subworkflows/varscan.cwl (this is the somatic varscan subWF) and def/tools/varscan_germline.cwl

susannasiebert:
@dufeiyu I believe we decided to drop the "workflow" from filenames. It is clear that this is a workflow by  virtue of it being in the subworkflows folder while the tool is in the tools folder. There shouldn't be any confusion about which is which. We have other instances of tools and subworkflows with the same name.

dufeiyu:
@susannasiebert Sure, actually my point is to add "varscan" in front of "germline": "varscan_germline", which is more specific than "germline_workflow.cwl".

PR #460, opened 2018-09-18
Add reference to new required file

johnegarza:


example_data/exome_workflow.yaml
```diff
@@ -57,3 +57,6 @@ varscan_strand_filter: 0
 vep_cache_dir: /gscmnt/gc2764/cad/ssiebert/toil_test/inputs/VEP_cache
 bqsr_intervals:
 - chr17
+dbsnp_vcf:
+  class: File
+  path: detect_variants/Homo_sapiens_assembly38.dbsnp138.vcf.gz
```
tmooney: I think we should alphabetize this (and the `bqsr_intervals` that was previously added) since all the other entries are sorted.  That'd make plain that there is an existing `dbsnp` entry in this YAML file as well!  Do we need both? Or should that have its key changed to `dbsnp_vcf` in lieu of this addition?

johnegarza: Good catch, I didn't realize this was just a rename and not a new parameter. I'll fix that now

PR #476, opened 2018-09-26
added minimal vep cache to example data

johnegarza:


johnegarza:
These changes were included in merged PR #487

PR #479, opened 2018-10-03
Add --tsl option to vep.cwl

susannasiebert:


susannasiebert:
I tried it locally and it will still run and add the field (TSL) but just not add any values for that field.

PR #483, opened 2018-10-12
Update strelka to 2.9.9; use standalone image

susannasiebert:


tmooney:
Ack! Just realized the helper name doesn’t match between the two images.

PR #491, opened 2018-10-31
Updating germline exome workflow to work with cromwell

apaul7:
Original CWL workflow was unable to successfully complete. Was an issue with finding the output file in the `definitions/tools/gatk_haplotype_caller.cwl` step. Unsure what the root cause was but with the help of @tmooney was able to find a workaround. 

Added an input for a `output_file_name` and moved the javascript expression to generate the name based on the input interval up to the `definitions/subworkflows/gatk_haplotypecaller_iterator.cwl` step.

 Also added small docker images.

### DIFF
--- a/definitions/tools/gatk_genotypegvcfs.cwl
+++ b/definitions/tools/gatk_genotypegvcfs.cwl
@@ -9,7 +9,7 @@ requirements:
       ramMin: 8000
     - class: InlineJavascriptRequirement
     - class: DockerRequirement
-      dockerPull: "mgibio/gatk-cwl:3.6.0"
+      dockerPull: "mgibio/gatk-cwl:3.5.0"
 arguments:
     ["-o", 'genotype.vcf.gz']
 inputs:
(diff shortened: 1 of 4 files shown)


definitions/tools/gatk_genotypegvcfs.cwl
```diff
@@ -3,13 +3,13 @@
...
 label: "GATK HaplotypeCaller"
-baseCommand: ["/usr/bin/java", "-Xmx8g", "-jar", "/opt/GenomeAnalysisTK-3.5.jar", "-T", "GenotypeGVCFs"]
+baseCommand: ["/usr/bin/java", "-Xmx8g", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "GenotypeGVCFs"]
 requirements:
     - class: ResourceRequirement
       ramMin: 8000
     - class: InlineJavascriptRequirement
     - class: DockerRequirement
-      dockerPull: mgibio/cle
+      dockerPull: "mgibio/gatk-cwl:3.6.0"
```
jasonwalker80: Should we be using GATK 3.5 here?

apaul7: waiting for one more +1 
https://github.com/genome/docker-gatk-cwl/pull/1

PR #492, opened 2018-11-01
Allow specifying names for CRAM files in somatic exome workflow.

tmooney:
The old default of both CRAMs having the same name is confusing and problematic, so here's some slightly better defaults too.


chrisamiller:
+1

PR #495, opened 2018-11-08
Base QC filenames on the CRAM filename.

tmooney:
This should help distinguish which QC matches which alignments, especially in cases where more than one CRAM is present (e.g. somatic pipelines).


zlskidmore:
+1

PR #496, opened 2018-11-13
Added a new optional custom vcf that incorporates clinvar data

johnegarza:


definitions/tools/vep.cwl
```diff
@@ -71,6 +71,20 @@ inputs:
...
                 }
             position: 6
+    custom_clinvar_vcf:
+        type: File?
+        secondaryFiles: [.tbi]
+        inputBinding:
+            valueFrom: |
+                ${
+                    if (inputs.custom_clinvar_vcf) {
+                        return ["--custom", inputs.custom_clinvar_vcf.path + "clinvar,vcf,exact,0,CLINSIGN,PHENOTYPE,SCORE,RCVACC,TESTEDINGTR,PHENOTYPELIST,NUMSUBMIT,GUIDELINES"]
```
tmooney: Does this need a separator in the string between the VCF path and the string "clinvar"?  Like:
```suggestion
                        return ["--custom", inputs.custom_clinvar_vcf.path + ",clinvar,vcf,exact,0,CLINSIGN,PHENOTYPE,SCORE,RCVACC,TESTEDINGTR,PHENOTYPELIST,NUMSUBMIT,GUIDELINES"]
```

johnegarza: Good catch, just fixed it

Issue #501, opened 2018-11-21
malformed requirements fields


chrisamiller:
There are a bunch of tools that have docker requirements slapped in like this: 
```
requirements:
    - class: DockerRequirement                                                                                                                                                                                                                                                      dockerPull: "mgibio/bisulfite"
arguments: [
    "--target", {valueFrom: "$(runtime.outdir)/trimmed_read"},
    "--threads", {valueFrom: "$(runtime.cores)"}
]
requirements:
    - class: ResourceRequirement
      ramMin: 16000
      tmpdirMin: 25000
      coresMin: 4
```

That's bad. Cromwell doesn't pick up on the docker requirements, and submits jobs with no `-a` or `-q` parameters, causing the jobs to PEND indefinitely.  

They need to be reformatted into a single block like this:
```
requirements:
    - class: ResourceRequirement
      ramMin: 16000
      tmpdirMin: 25000
      coresMin: 4
    - class: DockerRequirement
      dockerPull: "mgibio/bisulfite"
```

I'm taking care of the rnaseq-pipeline cwls, but we need a systematic look through to find the rest.

chrisamiller:
Did some searches and found no other CWLs with this issue.  I'm too lazy to trace back the origin, but I suspect it happened in a merge gone awry.  Closing!

PR #502, opened 2018-11-22
simplifying strand parameter

chrisamiller:
- Simplifies the strand parameter. Javascript expressions pass parameters based on strandedness. missing or nonsensical values assume unstranded libraries (should we assume first instead?)

- makes a few small updates to requirements fields to address 
https://github.com/genome/analysis-workflows/issues/501

### DIFF
--- a/definitions/pipelines/rnaseq.cwl
+++ b/definitions/pipelines/rnaseq.cwl
@@ -40,16 +40,12 @@ inputs:
        type: File
     gene_transcript_lookup_table:
        type: File
-    firststrand:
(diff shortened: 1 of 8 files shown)


tmooney:
Have you tried these changes out?  I'd expect them to need the `InlineJavascriptRequirement` (but some simple expressions don't).

chrisamiller:
Yep, they have all run successfully on small test bams.  I'm adding one more change related to mem usage here, since I just had some failures on full datasets.  update incoming in a sec...

PR #506, opened 2018-11-29
Adding in a cwl tool for cnvkit

johnegarza:


definitions/tools/cnvkit.cwl
```diff
@@ -0,0 +1,108 @@
...
+        type: File?
+        inputBinding:
+            position: 5
+            prefix: "--access"
+    method:
+        type: string?
+        inputBinding:
+            position: 6
+            prefix: "--method"
+            default: "hybrid"
```
tmooney: I believe the `default` goes one level higher rather than inside the `inputBinding`.

definitions/tools/cnvkit.cwl
```diff
@@ -0,0 +1,108 @@
+#! /usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+baseCommand: ["/usr/bin/python", "/usr/local/bin/cnvkit.py", "batch"]
+
+requirements:
+    - class: ShellCommandRequirement
+    - class: DockerRequirement
+      dockerPull: "etal/cnvkit"
```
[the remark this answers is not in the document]
johnegarza: Added that in

definitions/tools/cnvkit.cwl
johnegarza: Fixed!

definitions/tools/cnvkit.cwl
```diff
@@ -0,0 +1,108 @@
...
+        type: string?
+        default: "hybrid"
+        inputBinding:
+            position: 6
+            prefix: "--method"
+    threads:
+        type: int?
+        inputBinding:
+            position: 7
+            prefix: "-p"
```
tmooney: Should there be a corresponding resource requirement for the number of cores?  For example, here's how name sort specified this:
https://github.com/genome/analysis-workflows/blob/0a9a4ce83b49ed4e7eee5bcc09d83725136a36b0/definitions/tools/name_sort.cwl#L8

definitions/tools/cnvkit.cwl
```diff
@@ -0,0 +1,108 @@
...
+            glob: $(inputs.tumor_bam.nameroot).antitargetcoverage.cnn
+    tumor_target_coverage:
+        type: File
+        outputBinding:
+            glob: $(inputs.tumor_bam.nameroot).targetcoverage.cnn
+    tumor_bin_level_ratios:
+        type: File
+        outputBinding:
+            glob: $(inputs.tumor_bam.nameroot).cnr
+    tumor_segmented_rations:
```
chrisamiller: "rations" or "ratios"?

definitions/tools/cnvkit.cwl
johnegarza: There's no explicit flag from the program itself to set this. According to the documentation, 
`-p [PROCESSES], --processes [PROCESSES]
                        Number of subprocesses used to running each of the BAM
                        files in parallel. Without an argument, use the
                        maximum number of available CPUs. [Default: process
                        each BAM in serial]`

tmooney: If we're only running on a single tumor BAM, does this option need to be exposed at all?

If it does, I think the `coresMin` resource requirement should match the integer value here one way or another.

johnegarza: True, this may not be necessary here, since cromwell/cwl will take care of launching this multiple times as needed if we have multiple files. I'll remove the option.

definitions/tools/cnvkit.cwl
johnegarza: Fixed

johnegarza:
@chrisamiller That option is for `cnvkit segment`, and the one included in this file is `cnvkit batch`. I'm going to rename the tool to make this clear in case we decide to incorporate `segment` or any others later on.

chrisamiller:
Weird `batch` is just a wrapper and calls `segment` internally, so it seems like there oughta be a way to pass params down to the sub-steps.  Checking the code, though, there is not, so yeah, okay. I withdraw my comment!

PR #508, opened 2018-11-29
Added minimal clinvar vcf and index

johnegarza:
Include example data that can be used with #496


johnegarza:
Include example data that can be used with #496

PR #509, opened 2018-11-30
Added gatherer into the germline_exome pipeline

johnegarza:


definitions/pipelines/germline_exome.cwl
```diff
@@ -188,3 +193,35 @@ steps:
...
+                                function flatten(inArr, outArr) {
+                                    var arrLen = inArr.length;
+                                    for (var i = 0; i < arrLen; i++) {
+                                        if (Array.isArray(inArr[i])) {
+                                            flatten(inArr[i], outArr);
+                                        }
+                                        else {
+                                            outArr.push(inArr[i]);
+                                        }
+                                    }
```
tmooney: How about a `for...of`, like:
```javascript
for (let item of inArr) {
  if (Array.isArray(item)) {
    flatten(item, outArr);
  }
  else {
    outArr.push(item);
  }
}
```

definitions/pipelines/germline_exome.cwl
```diff
@@ -188,3 +193,35 @@ steps:
...
+                                    return outArr;
+                                }
+                                var no_secondaries = flatten(self, []);
+                                var all_files = []; 
+                                var arrLen = no_secondaries.length;
+                                for (var i = 0; i < arrLen; i++) {
+                                    all_files.push(no_secondaries[i]);
+                                    var secondaryLen = no_secondaries[i].secondaryFiles.length;
```
tmooney: Would a name like `primary_inputs` better convey what this is, instead of `no_secondaries` (which confused me when this then asks for its members `secondaryFiles` 😄)?

definitions/pipelines/germline_exome.cwl
```diff
@@ -188,3 +193,35 @@ steps:
...
         out:
             [gvcf, final_vcf, coding_vcf, limited_vcf, vep_summary]
+    gatherer:
+        run: ../tools/gatherer.cwl
+        in:
+            output_dir: output_dir
+            all_files:
+                source: [alignment_and_qc/per_target_hs_metrics, alignment_and_qc/hs_metrics, alignment_and_qc/per_base_hs_metrics, alignment_and_qc/alignment_summary_metrics, detect_variants/coding_vcf, detect_variants/gvcf, detect_variants/limited_vcf, detect_variants/vep_summary, detect_variants/final_vcf, alignment_and_qc/flagstats, alignment_and_qc/insert_size_metrics, alignment_and_qc/mark_duplicates_metrics, alignment_and_qc/per_base_coverage_metrics, alignment_and_qc/verify_bam_id_metrics, alignment_and_qc/verify_bam_id_depth, alignment_and_qc/insert_size_histogram, alignment_and_qc/per_target_coverage_metrics, alignment_and_qc/cram]
+                valueFrom: ${
+                                function flatten(inArr, outArr) {
```
tmooney: Maybe one day we'll get to use this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

It does also contain a suggested implementation called `flattenDeep` using `reduce` and `concat`.

PR #513, opened 2018-12-05
Upgrade pVACtools CWLs to latest version 1.1.5

susannasiebert:
This also fixes an incorrect reference to an input to pvacseq.cwl that I discovered buried in the travis-ci warnings.

### DIFF
--- a/definitions/pipelines/pvacseq.cwl
+++ b/definitions/pipelines/pvacseq.cwl
@@ -61,7 +61,7 @@ inputs:
         type: string?
     exclude_nas:
         type: boolean?
-    phased_proximal_variants_file:
+    phased_proximal_variants_vcf:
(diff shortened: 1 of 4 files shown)


definitions/pipelines/pvacseq.cwl
```diff
@@ -180,7 +180,7 @@ steps:
             fasta_size: fasta_size
             downstream_sequence_length: downstream_sequence_length
             exclude_nas: exclude_nas
-            phased_proximal_variants_file: phased_proximal_variants_file
+            phased_proximal_variants_vcf: phased_proximal_variants_file
```
tmooney: Should we change the input name for this pipeline, too?

(How did this mismatch slip through before?  Travis should notice this!)

PR #517, opened 2018-12-10
adding the Cell Ranger analysis cwls

apaul7:
I have the different CWLs I have been using to run different single cell cell ranger analysis. For `cellranger mkfastq`, `cellranger count`, `cellranger vdj`. This is currently using Eddie’s docker image for cellranger v2.2.0 which is hosted on the local private registry.

I also created a subworkflow that currently runs the `mkfastq` and `count` steps for a single sample. I have ran it using the tiny-bcl test data that can be downloaded from 10x Genomics website in about 15 minutes.

### DIFF
--- a/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
+++ b/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
@@ -6,15 +6,15 @@ label: "running cellranger mkfastq and count"
 
 inputs:
     bcl_directory:
-        type: string
+        type: Directory
     chemistry:
         type: string?
     reference:
-        type: string
+        type: Directory
     sample_name:
         type: string
     simple_sample_csv:
-        type: string
+        type: File
 
 steps:
     mkfastq:
--- a/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
+++ b/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
@@ -1,0 +1,48 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "running cellranger mkfastq and count"
+
+inputs:
+    bcl_directory:
+        type: string
+    chemistry:
+        type: string
+    reference:
+        type: string
+    sample_name:
+        type: string
+    simple_sample_csv:
+        type: string
+    unique_id:
+        type: string
+
+steps:
+    mkfastq:
+        run: ../tools/cellranger_mkfastq.cwl
+        in:
+            bcl_directory: bcl_directory
+            simple_sample_csv: simple_sample_csv
+            unique_id: unique_id
+        out: [samplesheet_csv, fastq_dir]
+    count:
+        run: ../tools/cellranger_count.cwl
+        in: 
+            chemistry: chemistry
+            fastq_directory: mkfastq/fastq_dir
+            reference: reference
+            sample_name: sample_name
+            unique_id: unique_id
+        out: [out_dir]
+
+outputs:
+    counts_out_dir:
+        type: Directory
+        outputSource: count/out_dir
+    fastqs:
+        type: Directory
+        outputSource: mkfastq/fastq_dir
+    samplesheet:
+        type: File
+        outputSource: mkfastq/samplesheet_csv
--- a/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
+++ b/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
@@ -15,8 +15,6 @@ inputs:
         type: string
     simple_sample_csv:
         type: string
-    unique_id:
-        type: string
 
 steps:
     mkfastq:
@@ -24,7 +22,6 @@ steps:
         in:
             bcl_directory: bcl_directory
             simple_sample_csv: simple_sample_csv
-            unique_id: unique_id
         out: [samplesheet_csv, fastq_dir]
     count:
         run: ../tools/cellranger_count.cwl
@@ -33,7 +30,6 @@ steps:
             fastq_directory: mkfastq/fastq_dir
             reference: reference
             sample_name: sample_name
-            unique_id: unique_id
         out: [out_dir]
 
 outputs:
--- a/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
+++ b/definitions/subworkflows/cellranger_mkfastq_and_count.cwl
@@ -8,7 +8,7 @@ inputs:
     bcl_directory:
         type: string
     chemistry:
-        type: string
+        type: string?
     reference:
         type: string
     sample_name:
--- a/definitions/tools/cellranger_count.cwl
+++ b/definitions/tools/cellranger_count.cwl
@@ -47,4 +47,4 @@ outputs:
     out_dir:
         type: Directory
         outputBinding:
-            glob: "$(inputs.unique_id)/outs/"
+            glob: "cellranger_output/outs/"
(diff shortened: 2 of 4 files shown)


definitions/subworkflows/cellranger_mkfastq_and_count.cwl
```diff
@@ -0,0 +1,48 @@
...
+
+cwlVersion: v1.0
+class: Workflow
+label: "running cellranger mkfastq and count"
+
+inputs:
+    bcl_directory:
+        type: string
+    chemistry:
+        type: string
```
chrisamiller: Since auto is the default, seems like this could be `string?`

definitions/subworkflows/cellranger_mkfastq_and_count.cwl
```diff
@@ -0,0 +1,48 @@
...
+            bcl_directory: bcl_directory
+            simple_sample_csv: simple_sample_csv
+            unique_id: unique_id
+        out: [samplesheet_csv, fastq_dir]
+    count:
+        run: ../tools/cellranger_count.cwl
+        in: 
+            chemistry: chemistry
+            fastq_directory: mkfastq/fastq_dir
+            reference: reference
```
chrisamiller: Do we have a directory with the weird cellranger ref structure set up?  (I know I have one, but we should get one into /core/ for at least b38 and mm10.   Also, what's the position on including example YAMLs with new workflows.  I'm gonna go ahead and say it'd be a good idea to add one to this PR, if you have one!

definitions/tools/cellranger_count.cwl
```diff
@@ -0,0 +1,57 @@
...
+            separate: false
+        doc: "Transcriptome reference compatible with input species and Cell Ranger"
+    sample_name:
+        type: string
+        inputBinding:
+            prefix: --sample=
+            position: 4
+            separate: false
+        doc: "Sample name, must be same as name specified in sample sheet in previous mkfastq step"
+    unique_id:
```
chrisamiller: As is (if I'm parsing this right), we'll always have to glob for `workflow_output/UNIQUEID/...` in the worfklow dir. Since workflows all get their own output directory anyway, is there any reason not to just hardcode this to be "cellranger_count" or something like that?  Let me know if I'm missing something...

definitions/tools/cellranger_vdj.cwl
```diff
@@ -0,0 +1,48 @@
...
+        type: string
+        inputBinding:
+            prefix: --id=
+            position: 4
+            separate: falsei
+        doc: "Unique name for the run"
+outputs:
+    out_dir:
+        type: Directory
+        glob: "$(inputs.unique_id)/outs/"]
```
johnegarza: Extra dangling ] seems to be the source of the checker failure

definitions/tools/cellranger_count.cwl
apaul7: Yes we could hard code it. I just liked having that option to allow you to easily know what's in the directory. What I've been doing is just adding the name of the tool to the end of the sample name. "sample1_count" or "sample1_vdj" and I realize now we can use the sample name input to create the unique id and remove the need for this in the yaml

definitions/subworkflows/cellranger_mkfastq_and_count.cwl
apaul7: Yup!

definitions/tools/cellranger_count.cwl
chrisamiller: I'm just thinking in terms of consistency across pipelines too.  The alignment, RNAseq, etc workflows don't have subdirs with sample names. They just contain the outputs.  That makes some operations easier: If I have a list of output dirs, it's easy to grab all of some result file without extra globbing.  We probably should get some feedback from the others and go with whatever the consensus is!

chrisamiller: required param, but could be set to something like "cellranger_output" so that it's consistent across samples.  We actually need to retain the whole directory and substructure, but it can be named whatever.

chrisamiller:
I'm ready to try these out once once they're merged. Let me know if I can be of help, @apaul7 !

definitions/tools/cellranger_count.cwl
```diff
@@ -0,0 +1,50 @@
...
+        inputBinding:
+            prefix: --sample=
+            position: 4
+            separate: false
+        doc: "Sample name, must be same as name specified in sample sheet in previous mkfastq step"
+outputs:
+    out_dir:
+        type: Directory
+        outputBinding:
+            glob: "$(inputs.unique_id)/outs/"
```
tmooney: `inputs.unique_id` in these two tools still needs to be updated to reflect aafbd19601f89281944385d0124c7758e9dda81d.

definitions/subworkflows/cellranger_mkfastq_and_count.cwl
```diff
@@ -0,0 +1,44 @@
...
+    bcl_directory:
+        type: string
+    chemistry:
+        type: string?
+    reference:
+        type: string
+    sample_name:
+        type: string
+    simple_sample_csv:
+        type: string
```
tmooney: Looks like we're going all-in on the "strings for filepaths" approach in this workflow.

apaul7:
I updated the inputs to work as a File or a Directory to allow it to work in other systems that may not have the shared file system as we do.
@chrisamiller For the references I have been using:
`/gscmnt/gc2134/finishing/var/reference/refdata-cellranger-GRCh38-1.2.0`
But it would be useful to have the different versions of 10x references for human/mouse in a central location. Not sure if I can/should do that?

Below is an example yaml I used to test out the subworkflow step. Took about 15 minutes to complete. the input `bcl_directory` and `simple_sample_csv` were downloaded from the 10x website, [here](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/2.2/using/mkfastq)

Also had to manually set the chemistry for this test because the `auto` chemistry mode maps the first 10,000 reads to determine the chemistry. Which causes the small test set to fail. But manually setting it allows it to run to completion. 
```
bcl_directory:
  class: Directory
  path: /gscmnt/gc2764/cad/alex.paul/scratch/cellranger/test/small-test/cellranger-tiny-bcl-1.2.0/
simple_sample_csv: 
  class: File
  path: /gscmnt/gc2764/cad/alex.paul/scratch/cellranger/test/small-test/cellranger-tiny-bcl-simple-1.2.0.csv 
sample_name: test_sample
reference: 
  class: Directory
  path: /gscmnt/gc2134/finishing/var/reference/refdata-cellranger-GRCh38-1.2.0/
chemistry: SC3Pv2
```

definitions/tools/cellranger_count.cwl
```diff
@@ -0,0 +1,50 @@
...
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Run Cell Ranger Count"
+
+baseCommand: ["cellranger", "count", "--localmem=64", "--localcores=8", "--id=cellranger_output"]
+
+requirements:
+    - class: DockerRequirement
+      dockerPull: "registry.gsc.wustl.edu/ebelter/cellranger:2.2.0"
```
chrisamiller: Can we update this to the 3.0 version that is now out?
`registry.gsc.wustl.edu/mgi/cellranger:3.0.1`

chrisamiller:
anything holding this up from being merged now (other than my request for updating to cellranger 3.0)?  Same commands should work fine, AFAIK

apaul7:
Was waiting on one more +1 after some of the changes I made

definitions/tools/cellranger_count.cwl
apaul7: I'll update the cwls to 3.0 and run my tests. Can make a new pr to do that so this one can get merged?

PR #521, opened 2018-12-12
Switching to version with needed cyvcf2 library

johnegarza:


johnegarza:
Added a PR to the repo for future reference, I'll follow the pattern Tom suggested next time: genome/docker-annotation_table-cwl#2

PR #531, opened 2019-01-07
Update vcf-readcount-annotator steps to use new -t option

susannasiebert:
This separates out the snv and indel outputs from bam-readcount and uses them separately to annotate the VCF using the new -t option of the vcf-readcount-annotator.

~Still missing is steps to left align and normalize the variants before running bam-readcount to ensure that snvs and indels are represented as separate VCF entries.~ DONE

I haven't run the restructured workflows but the tests are passing.


susannasiebert:
This is ready for code review.

susannasiebert:
I pushed another commit that removes the decompose steps before running bam-readcount by going back to just running `tools/bam_readcount.cwl` instead of the subworkflow. The decomposing is not needed there anymore since #549 adds a decompose step immediately after combine variants.

chrisamiller:
Is there a reason several of these workflows output bam_readcount tsvs, instead of just actually annotating the VCF?  I think I understand why we used to do it, but should we fix that?

chrisamiller:
oh and +1 on the decompose fixes

definitions/tools/bam_readcount.cwl
```diff
@@ -10,9 +10,7 @@ requirements:
     - class: DockerRequirement
       dockerPull: "mgibio/bam_readcount_helper-cwl:1.0.0"
 arguments: [
-    { valueFrom: $(runtime.outdir), position: -3 },
-    { valueFrom: " && ", shellQuote: false },
-    "/bin/cat", "$(runtime.outdir)/$(inputs.sample)_bam_readcount_snv.tsv", "$(runtime.outdir)/$(inputs.sample)_bam_readcount_indel.tsv"
```
tmooney: Looks like the `ShellCommandRequirement` can also be removed.

definitions/tools/vcf_readcount_annotator.cwl
```diff
@@ -7,7 +7,7 @@ label: "add bam_readcount info to vcf"
 baseCommand: ["vcf-readcount-annotator"]
 requirements:
     - class: DockerRequirement
-      dockerPull: "mgibio/vcf_annotation_tools-cwl:1.4.6"
+      dockerPull: "mgibio/vcf_annotation_tools-cwl:2.0.0"
```
tmooney: Should we go ahead and update to `3.0.0`?

susannasiebert: Done.

definitions/tools/bam_readcount.cwl
susannasiebert: Removed.

susannasiebert:
@tmooney I implemented the subworkflow you suggested. Could you have another look at that change at your convenience?

definitions/pipelines/detect_variants.cwl
```diff
@@ -259,47 +265,49 @@ steps:
...
         in:
             vcf: add_tumor_bam_readcount_to_vcf/annotated_bam_readcount_vcf
-            bam_readcount_tsv: normal_bam_readcount/bam_readcount_tsv
+            snv_bam_readcount_tsv: normal_bam_readcount/snv_bam_readcount_tsv
+            indel_bam_readcount_tsv: normal_bam_readcount/indel_bam_readcount_tsv
             data_type:
                 default: 'DNA'
             sample_name:
-                default: 'NORMAL'
+                default: 'TUMOR'
```
tmooney: This one should retain the default of `NORMAL`.

susannasiebert: Fixed.

susannasiebert:
@chrisamiller I'll pose our question to the larger group. For now I will merge this PR and if we decide to change what the files that get outputted by the various workflows we can make a new PR for that.

PR #533, opened 2019-01-09
Adding smoove.cwl to run Lumpy

apaul7:


definitions/tools/smoove.cwl
```diff
@@ -0,0 +1,48 @@
...
+        inputBinding:
+            prefix: --exclude
+            position: 4
+        doc: "Bed file used to exclude regions"
+
+outputs:
+    output_vcf:
+        type: File
+        outputBinding:
+            glob: "$(inputs.cohort_name)-smoove.genotyped.vcf.gz"
```
tmooney: Does smoove produce a tabix index with this that we'd want as a secondary file?

apaul7: yes. updating and testing it now

apaul7:
Would it be better to change the reference input to a string instead of a file so it can work with the current workflows?

tmooney:
We'll need to change it to a string if we add this step to one of our existing workflows, yes.

PR #534, opened 2019-01-11
General updates, and parameter modification for single sample runs

johnegarza:


definitions/tools/cnvkit_batch.cwl
```diff
@@ -8,24 +8,28 @@ requirements:
...
             prefix: "--normal"
     bait_intervals:
-        type: File
+        type: File?
         inputBinding:
             position: 3
             prefix: "--targets"
     reference:
-        type: string
+        type: string?
```
jasonwalker80: Is a reference FASTA optional?

johnegarza: It is if we decide to keep only one cnvkit batch cwl- this could be split into two, one for single sample and one for paired mode. It's optional now because in single sample mode, the only required arguments are the sample bam to be analyzed and a reference cnn file

PR #535, opened 2019-01-12
Specified more resource requirements/options for manta

johnegarza:


definitions/tools/manta_somatic.cwl
```diff
@@ -2,13 +2,17 @@
...
-label: "Set up and execute manta using helper script"
+label: "Set up and execute manta"
 
 requirements:
     - class: DockerRequirement
       dockerPull: mgibio/manta_somatic-cwl:1.5.0
     - class: InlineJavascriptRequirement
     - class: ShellCommandRequirement
+    - class: ResourceRequirement
+      coresMin: 2
```
jasonwalker80: Should we use a few more cores? In the GMS we were running with 12 and 24GB of RAM. I have not performed a benchmark but I presume there was a reason to use a large number of threads and RAM in the GMS.

definitions/tools/manta_somatic.cwl
```diff
@@ -48,6 +52,11 @@ inputs:
         inputBinding:
             position: -7
             prefix: "--outputContig"
+    jobs:
```
jasonwalker80: Do we want to define the number of jobs using an input parameter, or should we resolve the command line argument from the `ResourceRequirement`? Here is an example:
https://github.com/genome/analysis-workflows/blob/9ad798f18104eb4a9c6c118f42691b9f5d9efd19/definitions/tools/stringtie.cwl#L16

PR #536, opened 2019-01-15
adding subworkflow to call 3 sv callers

apaul7:
this should allow us to call smoove, manta and cnvkit for single sample sv analysis 
will need to add another subworkflow to run trios/cohorts 

Only required inputs is a cram file, string path to a reference, and a reference_cnn for cnvkit

### DIFF
--- a/definitions/subworkflows/single_sample_sv_callers.cwl
+++ b/definitions/subworkflows/single_sample_sv_callers.cwl
@@ -1,0 +1,141 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "Subworkflow to allow calling different SV callers which require bam files as inputs"
+
+requirements:
+  - class: InlineJavascriptRequirement
+
+inputs:
+    cram:
+        type: File
+    reference:
+        type: string
+
+    cnvkit_access:
+        type: File?
+    cnvkit_bait_intervals:
+        type: File?
+    cnvkit_diagram:
+        type: boolean?
+    cnvkit_drop_low_coverage:
+        type: boolean?
+    cnvkit_method:
+        type: string?
+    cnvkit_reference_cnn:
+        type: File?
+    cnvkit_scatter_plot:
+        type: boolean?
+
+    manta_call_regions:
+        type: File?
+    manta_non_wgs:
+        type: boolean?
+    manta_output_contigs:
+        type: boolean?
+
+    smoove_exclude_regions:
+        type: File?
+
+outputs:
+    intervals_antitarget:
+        type: File?
+        outputSource: run_cnvkit/intervals_antitarget
+    intervals_target:
+        type: File?
+        outputSource: run_cnvkit/intervals_target
+    normal_antitarget_coverage:
+        type: File?
+        outputSource: run_cnvkit/normal_antitarget_coverage
+    normal_target_coverage:
+        type: File?
+        outputSource: run_cnvkit/normal_target_coverage
+    reference_coverage:
+        type: File?
+        outputSource: run_cnvkit/reference_coverage
+    cn_diagram:
+        type: File?
+        outputSource: run_cnvkit/cn_diagram
+    cn_scatter_plot:
+        type: File?
+        outputSource: run_cnvkit/cn_scatter_plot
+    tumor_antitarget_coverage:
+        type: File
+        outputSource: run_cnvkit/tumor_antitarget_coverage
+    tumor_target_coverage:
+        type: File
+        outputSource: run_cnvkit/tumor_target_coverage
+    tumor_bin_level_ratios:
+        type: File
+        outputSource: run_cnvkit/tumor_bin_level_ratios
+    tumor_segmented_ratios:
+        type: File
+        outputSource: run_cnvkit/tumor_segmented_ratios
+    manta_diploid_variants:
+        type: File?
+        outputSource: run_manta/diploid_variants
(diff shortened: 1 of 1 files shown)


definitions/subworkflows/single_sample_sv_callers.cwl
```diff
@@ -0,0 +1,141 @@
...
+            non_wgs: manta_non_wgs
+            output_contigs: manta_output_contigs
+            reference: reference
+            tumor_bam: index_bam/indexed_bam
+        out: 
+            [diploid_variants, somatic_variants, all_candidates, small_candidates, tumor_only_variants]
+    run_smoove:
+        run: ../tools/smoove.cwl
+        in:
+            bams:
```
tmooney: Did you try specifying this more directly without the JS?  It seems like a variation on this ought to work: https://github.com/genome/analysis-workflows/blob/33c06a8096670ed8c291b79e3f3a3b2b8af80533/definitions/subworkflows/pindel_cat.cwl#L40-L41
(Not sure if we'd still need the `MultipleInputFeatureRequirement` even though there's only one source here.)

definitions/subworkflows/single_sample_sv_callers.cwl
```diff
@@ -0,0 +1,141 @@
...
+    manta_all_candidates:
+        type: File
+        outputSource: run_manta/all_candidates
+    manta_small_candidates:
+        type: File
+        outputSource: run_manta/small_candidates
+    manta_tumor_only_variants:
+        type: File?
+        outputSource: run_manta/tumor_only_variants
+    smoove_output_variant:
```
tmooney: Should this be called `smoove_output_variants` to match the other `*_variants` outputs?

apaul7: good catch. Just updated the output name

definitions/subworkflows/single_sample_sv_callers.cwl
```diff
@@ -0,0 +1,141 @@
...
+    run_cnvkit:
+        run: ../tools/cnvkit_batch.cwl
+        in:
+            access: cnvkit_access
+            bait_intervals: cnvkit_bait_intervals
+            diagram: cnvkit_diagram
+            drop_low_coverage: cnvkit_drop_low_coverage
+            method: cnvkit_method
+            reference_cnn: cnvkit_reference_cnn
+            tumor_bam: index_bam/indexed_bam
```
jasonwalker80: For Manta, I know we discussed running the tumor-only mode to use more sensitive parameters. This was required to identify a known SV. Do we have similar information to support running CNVkit in tumor-only mode? Is that why we are providing the single-sample input to the tumor_bam?

definitions/subworkflows/single_sample_sv_callers.cwl
apaul7: I tried
```
bams: [index_bam/indexed_bam]
```
But when I run `cwltool --validate` on the subworkflow it complains that a source of type `File` is incompatible with sink of type `{"type": "array", "items": "File:}`

Which led me to [this](https://www.biostars.org/p/218153/). And that's how I implemented it. Not sure if there is a better way now. 

I wrote the `smoove.cwl` to handle array of bams to allow the same tool work with single sample and groups. I could change that so that we have a `single_sample_smoove.cwl` and a `cohort_smoove.cwl` and then remove the javascript requirement.

definitions/subworkflows/single_sample_sv_callers.cwl
apaul7: My understanding of CNVkit is you're always comparing the tumor to a normal. There are 3 options for what you can use as the normal.
1) a paired normal bam
2) a generated reference.cnn (allowing pooling of normals)
3) "flat" coverage (don't need a normal)

Right now the CWLs do not fully reflect these requirements of running CNVkit. We will need to implement some exclusive and maybe dependent parameters to get the requirements of CNVkit in CWL. 
We are planning on using the reference.cnn which works in the current setup. And using a reference.cnn created by pooling normals together produces slightly better results than just running the matched tumor normal bams together.

definitions/subworkflows/single_sample_sv_callers.cwl
tmooney: Looks like `linkMerge: merge_nested` is what would make the `MultipleInputFeatureRequirement` version work:
https://www.commonwl.org/v1.0/Workflow.html#WorkflowStepInput

I tried it out, so here's apaul7/cancer-genomics-workflow#1 if you like it!

definitions/subworkflows/single_sample_sv_callers.cwl
jasonwalker80: Sounds good. Let's be sure to review this in person at the Wednesday 10am.

definitions/subworkflows/single_sample_sv_callers.cwl
apaul7: I was running the test data through and it failed.
```
No coercion defined from wom value(s) of type 'class 
wom.types.WomMaybePopulatedFileType$' to 
'Array[class wom.types.WomMaybePopulatedFileType$]'
```

Switched to `linkMerged: merge_flattened` and it looks like the inputs are being created correctly for smoove. Testing with the real test data to confirm.

apaul7: test data workflow succeeded with `merge_flattened`

PR #540, opened 2019-01-17
Allow for some overhead between Java and job memory requests.

tmooney:
Since 16000 MB < 16 GB, this was previously telling the JVM to use more memory than was available!


chrisamiller:
+1

PR #541, opened 2019-01-18
updating cellranger to 3.0.1

apaul7:
updating to the latest version of cellranger. `cellranger` is not added to the path in the new images so have to specify the installation location


chrisamiller:
+1  👍

PR #547, opened 2019-01-22
Update add_vep_fields_to_table.cwl to use the VAtools' new vep-annotation-reporter

susannasiebert:


definitions/tools/add_vep_fields_to_table.cwl
```diff
@@ -4,31 +4,30 @@ cwlVersion: v1.0
...
+            position: 1
     vep_fields:
         type:
             type: array
             items: string
         inputBinding:
             prefix: ""
-            itemSeparator: ","
+            itemSeparator: " "
             separate: false
```
tmooney: Not something you changed here, but the `prefix` and `separate` keys could be removed from this `inputBinding` entirely.

definitions/tools/add_vep_fields_to_table.cwl
```diff
@@ -4,31 +4,30 @@ cwlVersion: v1.0
...
-            position: 3
+            position: 2
+    tsv:
+        type: File
+        inputBinding:
+            prefix: "-t"
```
tmooney: Is there a reason the `tsv` input moved down here? (Is it meant to now have `position: 3`?)

susannasiebert: tsv is no longer required in the vep-annotation-reporter, hence the removal of the `position` inputBinding. The order reflects the one in the tool, with required arguments on top.

tmooney: If it's optional, should the type be `type: File?` instead?  It could still have a `position` specified for those times when the file is supplied.

susannasiebert: Yes, good catch. Why would we want the `position` inputBinding though, when the argument of the tool is not positional?

tmooney: It doesn't necessarily matter if the tool doesn't care. If it's required to be before the positional arguments I might want to make that explicit.  (There's a default position of `0` so it'll happen to be first in this circumstance.)

susannasiebert: I'm pretty sure the tool doesn't care.

PR #549, opened 2019-01-25
Decompose variants after combining VCFs from multiple callers

susannasiebert:


chrisamiller:
Should this PR also address the bam-readcount decompose step, which now becomes unnecessary?

+1

susannasiebert:
The still open https://github.com/genome/analysis-workflows/pull/531 has been updated with a commit that removes the bam-readcount decompose.

chrisamiller:
Got it! +1 again

PR #552, opened 2019-01-29
Vep everything flag

johnegarza:
This PR adds the --everything flag as a default argument to vep. Some of the flags it turns on require a cache to be present, so it's now a required parameter in vep.cwl and all upstream files. This also simplifies some of the argument construction logic within vep.cwl. The new flag also includes some AF flags that were previously only added if there was no custom gnomad vcf; since these are now always present, in order to avoid naming collisions within the vcf fields, fields added from the custom file are prefixed with "custom_gnomad"

### DIFF
--- a/definitions/tools/vep.cwl
+++ b/definitions/tools/vep.cwl
@@ -16,10 +16,9 @@ arguments:
     "--vcf",
     "--plugin", "Downstream",
     "--plugin", "Wildtype",
-    "--symbol",
     "--term", "SO",
     "--transcript_version",
-    "--tsl",
+    "--everything", 
     "-o", { valueFrom: $(runtime.outdir)/annotated.vcf }]
 inputs:
     vcf:
@@ -28,16 +27,11 @@ inputs:
             prefix: "-i"
             position: 1
     cache_dir:
-        type: string?
+        type: string
         inputBinding:
             valueFrom: |
                 ${
-                    if (inputs.cache_dir) {
-                        return ["--offline", "--cache", "--dir", inputs.cache_dir ]
-                    }
-                    else {
-                        return "--database"
-                    }
+                    return ["--offline", "--cache", "--dir", inputs.cache_dir ]
                 }
             position: 4
     synonyms_file:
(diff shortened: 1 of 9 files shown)


definitions/tools/vep.cwl
```diff
@@ -28,16 +26,11 @@ inputs:
...
         inputBinding:
             valueFrom: |
                 ${
-                    if (inputs.cache_dir) {
-                        return ["--offline", "--cache", "--dir", inputs.cache_dir ]
-                    }
-                    else {
-                        return "--database"
-                    }
+                    return ["--offline", "--cache", "--dir", inputs.cache_dir ]
```
tmooney: Now that this is required, we could move `--offline` and `--cache` to the arguments and make this a normal `prefix: "--dir"` parameter without the `valueFrom`.

definitions/tools/vep.cwl
```diff
@@ -68,15 +61,10 @@ inputs:
             valueFrom: |
                 ${
                     if (inputs.custom_gnomad_vcf) {
-                        return ['--check_existing', '--custom', inputs.custom_gnomad_vcf.path + ',gnomADe,vcf,exact,0,AF,AF_AFR,AF_AMR,AF_ASJ,AF_EAS,AF_FIN,AF_NFE,AF_OTH,AF_SAS']
+                        return ['--check_existing', '--custom', inputs.custom_gnomad_vcf.path + ',custom_gnomAD,vcf,exact,0,AF,AF_AFR,AF_AMR,AF_ASJ,AF_EAS,AF_FIN,AF_NFE,AF_OTH,AF_SAS']
```
tmooney: Does this means we need to change this filter?
https://github.com/genome/analysis-workflows/blob/de81bd20ce8829ef64146115902a10e4fe7bad0b/definitions/tools/filter_vcf_gnomADe_allele_freq.cwl#L31

jasonwalker80: I would suggest leaving the custom tag as `gnomADe`.

 1. we don't have to update the filter
 2. the string `gnomADe` denotes it's from the "exome" and `gnomADw` would be "Whole-genome"
 3. In general I'd suggest keeping VCF tags as short as possible. The header contains additional information including a description. As long as the tag is a unique string.

jasonwalker80:
Any additional updates on the PR or comments. I'd like to see this merged soon. We have several pipelines that would benefit from this code changes.

definitions/tools/vep.cwl
susannasiebert: I don't quite understand how the custom VCF works. Will this include the `genomADe` string the in the VEP tags or are the `AF` etc tags used plainly and would, thus, conflict with the AF tag written with the `--everything` flag.

definitions/tools/vep.cwl
johnegarza: Updated in the latest commit fdad71a

susannasiebert:
Right now the everything flag is an option. I think it should just be hard-coded. @jasonwalker80 thoughts?

jasonwalker80:
Yes, that was my understanding after the cancer-focused meeting last week. The `--everything` flag becomes the default behavior. No need to define it as an input. That's one of the main reasons for using this single flag. The GMS has trouble setting boolean values and hard-coding them in the YAML or `process_inputs.pl` is often overlooked unintentionally omitting critical annotation such as HGVS gene names.

PR #553, opened 2019-02-01
adding markdups, fixing pileup

chrisamiller:
This PR: 
- adds a bisulfite-strand aware duplicate marking procedure
- fixes the pileup script to reflect the changed params in the newer version of biscuit
- explicitly specifies the v1.3 tag since older versions of the container will not work with this cwl

### DIFF
--- a/definitions/tools/biscuit_markdup.cwl
+++ b/definitions/tools/biscuit_markdup.cwl
@@ -1,0 +1,28 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Biscuit dedup"
+baseCommand: ["/usr/bin/biscuit", "markdup"]
+requirements:
+    - class: ShellCommandRequirement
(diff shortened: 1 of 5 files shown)


definitions/tools/biscuit_markdup.cwl
```diff
@@ -0,0 +1,28 @@
...
+    - class: ShellCommandRequirement
+    - class: ResourceRequirement
+      ramMin: 16000
+    - class: DockerRequirement
+      dockerPull: "mgibio/bisulfite"
+
+arguments:
+    ["/dev/stdout",
+    { shellQuote: false, valueFrom: "|" },
+    "/usr/bin/sambamba", "sort", "-t", $(runtime.cores), "-m", "8G", "-o", "$(runtime.outdir)/markdup.bam", "/dev/stdin"
```
tmooney: Should we specify multiple CPUs in the resource requirement, or is `-t 1` fine?

PR #556, opened 2019-02-01
Added a subworkflow for cnvkit that returns a vcf

johnegarza:
By default, cnvkit outputs results in a .cns file; this subworkflow runs cnvkit and converts the .cns file to a vcf. Additionally, since the base `cnvkit_batch.cwl` tool has different use cases with different requirements, almost all of its input parameters/outputs are optional. Since we know the exact use case of the `single_sample_sv_callers` subworkflow, irrelevant parameters and nonexistent outputs were removed in upstream steps, and a necessary parameter was updated (previously listed as optional).

### DIFF
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -22,6 +22,7 @@ inputs:
         type: File
     cnvkit_vcf_name:
         type: string?
+        default: $(inputs.tumor_bam.nameroot).cnvkit.vcf
 outputs:
     cn_diagram:
         type: File?
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -1,0 +1,70 @@
+#! /usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "Subworkflow that runs cnvkit in single sample mode and returns a vcf file"
+
+inputs:
+    tumor_bam:
+        type: File
+    method:
+        type: string?
+        default: "hybrid"
+    diagram:
+        type: boolean?
+    scatter_plot:
+        type: boolean?
+    drop_low_coverage:
+        type: boolean?
+    male_reference:
+        type: boolean?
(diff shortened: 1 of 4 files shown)


definitions/subworkflows/cnvkit_single_sample.cwl
```diff
@@ -0,0 +1,70 @@
...
+    scatter_plot:
+        type: boolean?
+    drop_low_coverage:
+        type: boolean?
+    male_reference:
+        type: boolean?
+    reference_cnn:
+        type: File
+    cnvkit_vcf_name:
+        type: string?
```
tmooney: Could be useful to have a default of something like:
`$(inputs.tumor_bam.nameroot).cnvkit.vcf`

definitions/tools/cnvkit_vcf_export.cwl
```diff
@@ -0,0 +1,44 @@
...
+label: "Convert default cnvkit .cns output to standard vcf format"
+
+requirements:
+    - class: DockerRequirement
+      dockerPull: etal/cnvkit:0.9.5
+    - class: ShellCommandRequirement
+baseCommand: ["/usr/bin/python", "/usr/local/bin/cnvkit.py", "call"]
+arguments: [
+    { position: -1, valueFrom: $(inputs.male_reference), prefix: "-y" },
+    "-o", "adjusted.tumor.cns",
```
tmooney: Would it be useful to make this CNS file available as an output, too (even though this pipeline currently ignores it)?

johnegarza: Cnvkit call only adds one additional field to the file, entirely based on numbers in the original file, and this data will be included in the vcf anyway, so I chose to ignore it. If for some reason it's needed, the original cns is currently an output, and this adjusted cns can be generated from it very quickly.

definitions/subworkflows/cnvkit_single_sample.cwl
johnegarza: Good idea, added in ` [957435e](/genome/analysis-workflows/pull/556/commits/957435e34fd7f3c50a6289146797522c22531f8a)`

PR #557, opened 2019-02-02
fixing cores/mem requirements, bisulfite data

chrisamiller:
These were meant to be two different PRs, but git happened ¯\_(ツ)_/¯

first is adding cores and updating mem requirements for the sambamba sort

second part is dropping in an example yaml and test bams for the bisulfite workflow.  I'd like to update those paths eventually, but a non-perfect yaml is better than no yaml.


definitions/tools/biscuit_markdup.cwl
```diff
@@ -8,13 +8,14 @@ requirements:
...
+      coresMin: 4
     - class: DockerRequirement
-      dockerPull: "mgibio/bisulfite"
+      dockerPull: "mgibio/bisulfite:v1.3"
 
 arguments:
     ["/dev/stdout",
     { shellQuote: false, valueFrom: "|" },
-    "/usr/bin/sambamba", "sort", "-t", $(runtime.cores), "-m", "8G", "-o", "$(runtime.outdir)/markdup.bam", "/dev/stdin"
+    "/usr/bin/sambamba", "sort", "-t", $(runtime.cores), "-m", "15G", "-o", "$(runtime.outdir)/markdup.bam", "/dev/stdin"
```
tmooney: How much memory does `biscuit` itself need?  Is  the remaining ~1GB enough?

chrisamiller: not a ton, but to be safe, I just pushed a fix that upped the overall memory to 24G.  As soon as the checks clear, I'll merge.  Thanks!

PR #558, opened 2019-02-02
passing ref param to bsvcf2bed

chrisamiller:
With ditching the perl script in favor of the built-in utility, we now have to pass the reference fasta to this tool.


chrisamiller:
It's already there:

      dockerPull: "mgibio/bisulfite:v1.3"

tmooney:
But aren't the changes this reflects new since after version 1.3?
https://github.com/genome/docker-bisulfite/compare/v1.3...79e47e42bb1d33865e5756e19df1572f65782d8d

PR #560, opened 2019-02-06
Add annotated VCF as an output to pvacseq.cwl workflow

susannasiebert:


susannasiebert:
@tmooney I just pushed another commit to this PR to implement a request @malachig made for creating a TSV from the annotated VCF. If you could review that as well, that would be greatly appreciated.

PR #564, opened 2019-02-08
Add manta somatic wf

johnegarza:


definitions/pipelines/somatic_exome.cwl
```diff
@@ -126,6 +126,13 @@ inputs:
     custom_clinvar_vcf:
         type: File?
         secondaryFiles: [.tbi]
+    manta_call_regions:
+        type: File?
+        secondaryFiles: [.tbi]
+    manta_non_wgs:
+        type: boolean?
```
tmooney: Should this have a default of `true` in this exome pipeline?

johnegarza: Good call, added in the latest commit

PR #566, opened 2019-02-11
Some updates to pVACseq CWLs

susannasiebert:


definitions/tools/pvacseq.cwl
```diff
@@ -20,6 +20,7 @@ arguments:
 inputs:
     input_file:
```
tmooney: Not something changed in this PR, but would a more descriptive name or a `doc` line to explain what sort of file this is be useful?

susannasiebert: I just pushed a commit to the PR to rename this input to `input_vcf`.

definitions/tools/pvacseq.cwl
```diff
@@ -4,22 +4,28 @@ cwlVersion: v1.0
...
-baseCommand: ["pvacseq", "run"]
+baseCommand: [
+    "ln", "-s"
+]
+arguments: [
+    { valueFrom: "$TMPDIR", shellQuote: false },
+    "/tmp/pvacseq",
+    { valueFrom: " && ", shellQuote: false },
+    "export", "TMPDIR=/tmp/pvacseq",
+    { valueFrom: " && ", shellQuote: false },
```
chrisamiller: This seems kind of weird.  I'm guessing that the issue is TMPDIR is set in a non-standard way within the cromwell job?

susannasiebert: Cromwell sets a unique tmpdir for each step that can get very long. Python multi threading will throw an error`AF_UNIX path too long` in that situation.  We think this is thrown only in pVACtools because it uses multithreading via  pymp.

susannasiebert: See
https://github.com/broadinstitute/cromwell/issues/3647

chrisamiller: Ugh. yeah, okay, I get it then, and don't see a more elegant way to do it than what you have here!  +1

susannasiebert:
@tmooney @chrisamiller I made a few more commits but I think this PR is now ready for final review before merging.

definitions/tools/pvacvector.cwl
```diff
@@ -5,19 +5,24 @@ class: CommandLineTool
...
+    "/tmp/pvacseq",
+    { valueFrom: " && ", shellQuote: false },
+    "export", "TMPDIR=/tmp/pvacseq",
+    { valueFrom: " && ", shellQuote: false },
+    "/opt/conda/bin/pvacvector",
     "run"
+    "--iedb-install-directory", "/opt/iedb",
+    { position: 5, valueFrom: $(runtime.outdir) },
 ]
 requirements:
```
tmooney: This one needs a `ShellCommandRequirement`, too!
```suggestion
requirements:
    - class: ShellCommandRequirement
```

PR #570, opened 2019-02-14
New VCF filter - somatic_llr

chrisamiller:
Adds a filter, as well as cleans up a few other things in the somatic_exome test yaml. It has passed the small chr17 test, but has not been tested in full on a real bam


definitions/tools/filter_vcf_somatic_llr.cwl
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "use the binomial/llr somatic filter to weed out low confidence variants"
+requirements:
+    - class: DockerRequirement
+      dockerPull: mgibio/somatic-llr-filter
```
tmooney: We might want to version this image.

definitions/tools/filter_vcf_somatic_llr.cwl
```diff
@@ -0,0 +1,27 @@
...
+    threshold: 
+         type: float
+         inputBinding:
+             prefix: "--llr-threshold"
+             position: -2
+outputs:
+     somatic_llr_filtered_vcf:
+         type: File
+         outputBinding:
+             glob: "somatic_llr_filtered.vcf"
```
tmooney: The formerly final filter yielded a generic name:
https://github.com/genome/analysis-workflows/blob/44ada20f3eeb59005d5bd999d2435102e9bae991/definitions/tools/filter_vcf_cle.cwl#L23

I don't have great suggestions on what to do for this one, but it seems like we may want to take a pass at all our naming through variant detection and filtration.

definitions/tools/filter_vcf_somatic_llr.cwl
chrisamiller: added a version

definitions/tools/filter_vcf_somatic_llr.cwl
chrisamiller: Yeah, I'm going to punt on it for this PR, and make an issue suggesting that we do some name harmonization

Issue #486, opened 2018-10-23
Missing VEP input `--assembly`, required when multiple genome assemblies exist.


jasonwalker80:
When multiple cache directories exist for the same Ensembl version, but different genome assemblies, ex. GRCh37 and GRCh38, exist, VEP exits with the following error message:
`ERROR: Multiple assemblies found for cache version 93 (GRCh37, GRCh38) - specify one using --assembly [assembly]`

https://github.com/genome/analysis-workflows/blob/13a7ce79bde922cd3ce260b1ab6d92f160602794/definitions/tools/vep.cwl#L25

PR #572, opened 2019-02-14
Added assembly flag to vep and included the parameter in upstream wor…

johnegarza:
…kflows


johnegarza:
This addresses #486. Thoughts on using GRCh38 as the default welcome

chrisamiller:
Would it be safer to have there be no default, and make it a required input? It always worries me that someone could be running a mouse build and end up with GRCh38 just because they failed to override the default

PR #574, opened 2019-02-15
Use a `long` for the genome size instead of an `int`.

tmooney:
Genome sizes exceed 32-bit integers. They don't seem *that* long 😄

Guess we've been spoiled by friendlier languages that don't cap their integers.  It's all really a string anyway.


chrisamiller:
+1

PR #578, opened 2019-02-18
versioning this, since the next image will change it

chrisamiller:


chrisamiller:
I also edited this into `/gscmnt/gc2560/core/processing-profile/cwl-pipeline/31e873b623e2454e9b68a53dac9356c4/tools/somatic_docm_filter.cwl` so the current default PP doesn't break when we push changes

chrisamiller:
This got resolved in a previous PR where I versioned everything.  Closing

PR #581, opened 2019-02-20
updating to stable cle docker version

chrisamiller:
locking things in at v1.3.1


chrisamiller:
This docker image is identical to the old "latest" with two exceptions (adding missing dependencies for cyvcf). This has been tested with the somatic_exome testing cwl as a sanity check and a cromwell build succeeds

PR #582, opened 2019-02-20
Bump up ram to get around errors

johnegarza:
We continue to see pindel jobs dying due to memory errors. Increasing requested memory again to 64 gb; there may be better solutions worth investigating, but for now, this seems to fix the issues.


chrisamiller:
+1

PR #590, opened 2019-02-27
Adding survivor merge for sv vcfs

apaul7:
this will allow us to merge the results from different sv callers together into a single vcf. And then annotate them with a population frequency of the sv calls.

I have it set up as a separate sub workflow to make it easier for people to run this manually. 

Open to suggestions on naming of files and input variables.

### DIFF
--- a/definitions/tools/add_sv_population_frequency.cwl
+++ b/definitions/tools/add_sv_population_frequency.cwl
@@ -19,7 +19,7 @@ requirements:
 inputs:
     vcf:
         type: File
-        doc: "vcf to add populationa allele frequence to"
+        doc: "vcf to add population allele frequencies to"
     sv_db:
         type: File
         doc: "bed file containing allele frequencies for a population"
--- a/definitions/tools/add_sv_population_frequency.cwl
+++ b/definitions/tools/add_sv_population_frequency.cwl
@@ -1,3 +1,5 @@
+#!/usr/bin/env cwl-runner
+
 cwlVersion: v1.0
 class: CommandLineTool
 label: "add population allele frequencies to a vcf"
--- a/definitions/tools/add_sv_population_frequency.cwl
+++ b/definitions/tools/add_sv_population_frequency.cwl
@@ -1,0 +1,32 @@
+cwlVersion: v1.0
+class: CommandLineTool
+label: "add population allele frequencies to a vcf"
+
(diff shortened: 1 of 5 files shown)


definitions/tools/add_sv_population_frequency.cwl
```diff
@@ -0,0 +1,32 @@
...
+    - class: DockerRequirement
+      dockerPull: "halllab/svtools:v0.4.0-2dc7f91" 
+    - class: ResourceRequirement
+      ramMin: 5000
+      coresMin: 1
+
+inputs:
+    vcf:
+        type: File
+        doc: "vcf to add populationa allele frequence to"
```
tmooney:
```suggestion
        doc: "vcf to add population allele frequencies to"
```

definitions/tools/survivor.cwl
```diff
@@ -0,0 +1,65 @@
...
+    same_strand:
+        type: int
+        inputBinding:
+            position: 5
+        doc: "Require merged SVs to be on the same strand, 1=yes"
+    estimate_sv_distance:
+        type: int
+        inputBinding:
+            position: 6
+        doc: "Estimate distance based on the sixe of SV, 1=yes"
```
tmooney:
```suggestion
        doc: "Estimate distance based on the size of SV, 1=yes"
```

definitions/tools/survivor.cwl
```diff
@@ -0,0 +1,65 @@
...
+    minimum_sv_calls:
+        type: int
+        inputBinding:
+            position: 3
+        doc: "Minimum number of sv calls needed to be merged"
+    same_type:
+        type: int
+        inputBinding:
+            position: 4
+        doc: "Require merged SVs to be of the same type, 1=yes"
```
tmooney: Is anything else `no`?

apaul7: it's 1=yes, 0=no. updated documentation to state that. I was going to use the boolean input type but thought that was for flag inputs. this has to have something as all inputs are positional

tmooney: Yeah, the boolean type is mostly for flag inputs.  You could use JavaScript to convert it to a `1` or a `0`, or possibly define an enumerated type here that accepts only `0` and `1`.

apaul7:
will merge this and make a new pr to add it to the single sample sv workflow

PR #591, opened 2019-02-28
Refactor DoCM workflow

chrisamiller:
This refactors the docm step in several ways:
- adds docm variant hotspots in after somatic callers have already been merged
- makes passing them optional - by default DoCM variants are merged into the final VCF, but flagged as failing by default. They can be kept as passing variants by setting `filter_docm_variants` to 0 in the input yaml.  

Some notes:
- there is still a line in filter_vcf_docm.cwl that refers to chrisamiller/cle  After the following PR is merged, and a new tagged version of mgibio/cle is built, it will need to be updated https://github.com/genome/docker-cle/pull/89
- This has not been tested on anything other than the small chr17 test data set (which contains no docm-passing variants).  A larger test will be necessary before this goes live to production.


definitions/pipelines/detect_variants.cwl
```diff
@@ -133,9 +136,6 @@ outputs:
         type: File
         outputSource: pindel/filtered_vcf
         secondaryFiles: [.tbi]
-    docm_unfiltered_vcf:
-        type: File
-        outputSource: docm/unfiltered_vcf
```
chrisamiller: I have verified with dave and feiyu that this raw output is not needed

definitions/pipelines/detect_variants.cwl
```diff
@@ -62,6 +62,9 @@ inputs:
     docm_vcf:
         type: File
         secondaryFiles: [.tbi]
+    filter_docm_variants:
+        type: int?
```
tmooney: If this is being used as a boolean, it'd be better to either have it be a boolean in the workflow (and translated to `1`/`0` in the tool) or to be an enum so that `0` and `1` are the only allowed values.

definitions/tools/docm_add_variants.cwl
```diff
@@ -0,0 +1,40 @@
...
+requirements:
+    - class: ResourceRequirement
+      ramMin: 9000
+      tmpdirMin: 25000
+    - class: DockerRequirement
+      dockerPull: mgibio/cle
+arguments:
+    ["-genotypeMergeOptions", "PRIORITIZE",
+     "--rod_priority_list", "callers,docm",
+     "--setKey","null",
```
dufeiyu: This is the second combine_variant set. So I assume those "set=<real_variant_callers>" in INFO field from the first combine_variant step will be preserved. But "null" here will not add "docm" to "set=". So what if docm variants is also called by real_callers ? There won't be something like "set=<real_caller>,docm". If the variant is only called by docm caller, it can probably be identified by INFO field without "set=".

chrisamiller: It's a good point, but if a docm variant is called by a real caller, I can't think of a reason why we'd need to know that.  The point of docm is to "rescue" things that weren't called by the standard callers, in my mind.  Does this clash with some of the existing CLE downstream workflow?

chrisamiller:
@dufeiyu One of the output files that will be in the final directory is the `docm/filtered_vcf`.  That's before it ever gets merged in with the rest (or gets annotated, etc) Is there any reason why that wouldn't meet your docm needs?

dufeiyu:
Dave wants docm variants also get vep annotation and bam readcount run, and are included in final annotated vcf.

chrisamiller:
We verified with Dave that this approach is fine for the CLE. We're debugging some failures on larger tests and will likely have some minor bugfixes before this can go live

johnegarza:
@chrisamiller Any updates? We're planning to make a new processing profile soon and would like to include this PR

chrisamiller:
I have screwed up this branch royally through an ill-advised series of git commands 
(basically, this: https://imgur.com/ccaTs94.gif)

I have ported all of the relevant changes into a new branch and a new pull-request:  
https://github.com/genome/analysis-workflows/pull/607

PR #596, opened 2019-03-06
Hacked in default vcf name

johnegarza:
The `default` field does not seem to allow parameter references/Javascript, so I made a workaround using the `valueFrom` field instead


tmooney:
Based on reading common-workflow-language/common-workflow-language/issues/377 I would expect that the output `glob` also needs to be changed.  (Any thoughts on using empty string versus the word `default` as here?  It seems unlikely that someone would want to name the file `default`, but using an invalid filename as the default might be better.)

PR #599, opened 2019-03-09
WIP: PP 92e81574f6ae4dfb8629c72aa765846d

jasonwalker80:
Save Grouped UMI alignments and use for Coverage QC before Consensue Calling.

This is being tested on internal AnP: 7ac3e101e1bf4ef4abd6684d18abebab

PR #600, opened 2019-03-10
New Filter Consensus Params for Single-Molecule

jasonwalker80:
The duplex parameters are way too strict for single-molecule consensus calling. Most of the UMI sequencing we do is not designed for consensus calling at all. Rather than ignore the UMIs, make single-molecule consensus reads but relax the parameters to allow nearly all consensus reads of reasonable quality pass.

This should replace PR #599 since the raw alignments are not really needed. 

NOTE: keeping the original alignments is not the right approach unless we implemented UMI-aware duplicate marking. The net effect is similar to this single-molecule consensus calling with relaxed params.

### DIFF
--- a/definitions/subworkflows/duplex_alignment.cwl
+++ b/definitions/subworkflows/duplex_alignment.cwl
@@ -17,6 +17,21 @@ inputs:
         type: string
     target_intervals:
        type: File?
+    min_reads:
+       type: int
+       default: [10, 5, 3]
+    max_read_error_rate:
+       type: float?
+       default: 0.05
+    max_base_error_rate:
+       type: float?
+       default: 0.1
+    min_base_quality:
+       type: int
+       default: 50
+    max_no_call_fraction:
+       type: float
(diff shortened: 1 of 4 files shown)


chrisamiller:
Seems reasonable! +1

jasonwalker80:
Correct, RE: `0.5` vs `0.05` for `max_no_call_fraction`. The higher value, 50%, would allow half of the single-molecule consensus to be `N` characters. This is high, but we are losing most of the consensus reads with the more strict 5% parameter. I was effectively turning off this filter but it's a required parameter. 50% seemed reasonable for a 150bp read length. That would leave at least 75bp of sequence data. Given that many of the read-lengths for degraded FFPE samples do fall in this range, this value was my best estimate.

jasonwalker80:
A reminder this is only for the single-molecule consensus we are using for Whole Exome Sequencing. The duplex processing profile should still be used for small, targeted panels with high depth coverage.

PR #603, opened 2019-03-11
Moved logic up a level to the subworkflow, so the tool doesn't need a…

johnegarza:
…ny expressions

### DIFF
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -70,15 +70,16 @@ steps:
             cnr_file: cnvkit_main/tumor_bin_level_ratios
             output_name: 
                 source: cnvkit_vcf_name
-                default: "default"
+                default: "."
                 valueFrom: |
                     ${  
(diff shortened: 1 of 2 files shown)


definitions/subworkflows/cnvkit_single_sample.cwl
```diff
@@ -65,7 +68,17 @@ steps:
...
             male_reference: male_reference
             cnr_file: cnvkit_main/tumor_bin_level_ratios
-            output_name: cnvkit_vcf_name
-            tumor_bam: tumor_bam
+            output_name: 
+                source: cnvkit_vcf_name
+                default: "default"
+                valueFrom: |
+                    ${  
+                        if(inputs.output_name != "default") {
```
tmooney: Could this check if `inputs.cnvkit_vcf_name` is defined and avoid needing the `default: "default"` altogether?

johnegarza: Testing a few different things right now. I'm pretty sure there was a reason I didn't do this the first time around, but I don't remember what the exact issue was, so I'll check again

PR #607, opened 2019-03-18
Retrying the docm workflow PR

chrisamiller:
Recreates https://github.com/genome/analysis-workflows/pull/591  after a series of adventures in git gone wrong.  Also addresses the Int/boolean issue, which was the last unresolved item from that PR.  

Still working on verifying this on a sample that may contain actual docm variants to be included, but that's running now


definitions/tools/docm_add_variants.cwl
```diff
@@ -0,0 +1,39 @@
...
+baseCommand: ["/usr/bin/java", "-Xmx8g", "-jar", "/opt/GenomeAnalysisTK.jar", "-T", "CombineVariants"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 9000
+      tmpdirMin: 25000
+    - class: DockerRequirement
+      dockerPull: mgibio/cle
+arguments:
+    ["-genotypeMergeOptions", "PRIORITIZE",
+     "--rod_priority_list", "callers,docm",
```
jasonwalker80: Does this mean we lose the list of callers? I know in several projects where the number of callers, ex. 3 of 4, was used as a post-pipeline filter.

definitions/subworkflows/docm_cle.cwl
```diff
@@ -37,24 +36,43 @@ steps:
...
+        in:
+            vcf: bgzip/bgzipped_file
+        out:
+            [indexed_vcf]
+    decompose:
+        run: ../tools/vt_decompose.cwl
+        in:
+            vcf: index/indexed_vcf
+        out:
+            [decomposed_vcf]
```
jasonwalker80: In the detect_variants.cwl workflow an index step was required. Is one not required here?

definitions/tools/docm_add_variants.cwl
chrisamiller: Nope!  see /gscmnt/gc2708/info/medseq/tmp/toiltest/output3/cromwell-executions/detect_variants.cwl/ab14a12b-b095-42e9-af4d-afa00cbca514/call-annotated_filter_index/execution/annotated_filtered.vcf.gz for an example

definitions/subworkflows/docm_cle.cwl
chrisamiller: it's indexed again after the filter is applied, below.

definitions/tools/docm_add_variants.cwl
jasonwalker80: awesome! thanks!

chrisamiller: to further expound - docm variants will never overwrite existing variants, only get added where there is no variant,  so the original line (with all caller info) will remain.

definitions/tools/filter_vcf_docm.cwl
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Filter variants from the DoCM detector"
+baseCommand: ["/usr/bin/perl", "/usr/bin/docm_filter.pl"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "mgibio/cle:v1.4.1"
+arguments: [
```
dufeiyu: $docm_vcf, $normal_cram, $tumor_cram, $output_vcf_file, $set_filter_flag is the order of arguments that docm_filter.pl takes. Just wanna make sure that "docm_filtered_variants.vcf" will be in the position after tumor_bam and before filter_docm_variants flag.

chrisamiller:
> Are those variants detected only by docm present ok in the final filtered vcf in your test ?
It depends.  They do appear in the docm output file, and then two of them make it through to the final filtered VCF (labelled as DOCM_ONLY  if `filter_docm_variants: true`). Several others are hard-filtered by other filters (gnomad, etc) and never make it to the very end.  If we set `filter_docm_variants: true`, they make it all the way through as PASS, again, unless they are removed by other filters.

definitions/tools/filter_vcf_docm.cwl
chrisamiller: Yep.  The negative numbers go before the `arguments` block, and positive numbers go after.

definitions/tools/docm_add_variants.cwl
```diff
@@ -0,0 +1,39 @@
...
+     "--setKey","null",
```
dufeiyu: Just wanna confirm that with this setting the variants only detected by docm will NOT have "set=" in INFO field

chrisamiller: Correct. there won't be any "set", because they weren't detected by any caller.

PR #608, opened 2019-03-18
updating version - handles missing AD/DP fields

chrisamiller:


chrisamiller:
This commit shows the changes:  https://github.com/genome/docker-somatic-llr-filter/commit/97b1e7dff44662cb557793f9b4db73623f43e8d9

chrisamiller:
Yeah, that was a mistake, and then I just rolled with it

PR #610, opened 2019-03-19
Simple Exome Alignment + QC

jasonwalker80:
A new basic workflow for generic, non-human exome alignment and qc.


definitions/pipelines/simple_exome_alignment.cwl
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "simple xome alignment with qc"
```
tmooney:
```suggestion
label: "simple exome alignment with qc"
```

jasonwalker80:
This is replaced by #713

PR #612, opened 2019-03-28
adding depth filter

chrisamiller:
I thought I pushed this yesterday, but apparently did not.  In any case...

### DIFF
--- a/definitions/tools/filter_vcf_depth.cwl
+++ b/definitions/tools/filter_vcf_depth.cwl
@@ -1,0 +1,28 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "filter variants at sites below a given sequence depth in each sample"
+requirements:
+    - class: DockerRequirement
+      dockerPull: mgibio/depth-filter:0.1
+    - class: ResourceRequirement
+      ramMin: 4000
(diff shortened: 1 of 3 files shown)


definitions/tools/filter_vcf_depth.cwl
```diff
@@ -0,0 +1,28 @@
...
+class: CommandLineTool
+label: "filter variants at sites below a given sequence depth in each sample"
+requirements:
+    - class: DockerRequirement
+      dockerPull: mgibio/depth-filter:0.1
+    - class: ResourceRequirement
+      ramMin: 4000
+baseCommand: ["/opt/conda/bin/python3","/usr/bin/depth_filter.py"]
+arguments: 
+    ["TUMOR,NORMAL", {valueFrom: "$(runtime.outdir)/depth_filtered.vcf"}]
```
tmooney: It might make sense to parameterize the sample names in the tool and then specify this value in workflows that call it. Something like this:
https://github.com/genome/analysis-workflows/blob/60d8a9e6c5f571ec9b37f10290a1f4613013f3e1/definitions/pipelines/detect_variants.cwl#L279-L280

chrisamiller: Good call - added that, and pushed the dv changes that I forgot to commit last night

chrisamiller:
Yep- it's agnostic to order, just drops a site if either is below the threshold

PR #617, opened 2019-04-03
Parameterizing vep plugins

apaul7:
allows workflows to pick and choose vep plugins to run. 
currently have the same defaults set for all workflows but a future PR will change the plugins used in the germline workflow

### DIFF
--- a/definitions/tools/vep.cwl
+++ b/definitions/tools/vep.cwl
@@ -16,8 +16,6 @@ arguments:
     ["--format", "vcf",
     "--vcf",
     "--fork", "4",
-    "--plugin", "Downstream",
-    "--plugin", "Wildtype",
     "--term", "SO",
     "--transcript_version",
     "--offline",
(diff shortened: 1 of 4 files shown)


definitions/tools/vep.cwl
```diff
@@ -90,6 +88,10 @@ inputs:
         inputBinding:
             prefix: "--fasta" 
             position: 8
+    plugins:
+        type: string[]?
+        inputBinding:
+            prefix: "--plugin"
```
tmooney: To get the `--plugin` prefix before each plugin, this will need to be something like:
```cwl
    type:
      type: array
      items: string
      inputBinding:
        prefix: "--plugin"
```

See also [this page](https://www.commonwl.org/user_guide/09-array-inputs/) for examples of three different ways to handle array inputs.

apaul7: Good catch!

PR #618, opened 2019-04-03
Update interval input

jasonwalker80:
Something is wrong. this PR was not intended to have commit 5f6cf43. I'll have to close this and make a new one. For now, WIP.


tmooney:
(Of course, my reply is about the part you said shouldn't be in here :smile:)

PR #619, opened 2019-04-04
use sample name as run id

apaul7:
Cellranger has an input option for a run id which is used to generate the running directory and is used as an identifier in some of the output files. Initially we had a generic `cellranger_ouput` id to simplify parsing  and consistency across runs. That makes it difficult to easily determine the sample from the result files. This PR helps to fix that.


chrisamiller:
+1 looks good!

malachig:
https://xkcd.com/327/

PR #622, opened 2019-04-05
Updated vep to version 95; plugins remain unchanged

johnegarza:


susannasiebert:
I've been a bit out of the loop here but would it make sense to switch the versioning from 1.1.0 to the actual VEP version so that it's easier to keep track of what is being run? (this is obviously something that would need to change in the appropriate docker repo)

PR #626, opened 2019-04-16
adding cellranger-atac count cwl

apaul7:
this pr adds a cwl that can be used to run cellranger-atac count. I'll end up using this to create a processing_profile to handle single cell atac samples in the GMS.

### DIFF
--- a/definitions/tools/cellranger_atac_count.cwl
+++ b/definitions/tools/cellranger_atac_count.cwl
@@ -4,8 +4,8 @@ cwlVersion: v1.0
 class: CommandLineTool
 label: "Run Cell Ranger ATAC Count"
 
-baseCommand: ["/opt/cellranger-atac-1.0.1/cellranger-atac", "count", "--localmem=64", "--localcores=8"]
-arguments: ["--id=$(inputs.sample_name)"]
+baseCommand: ["/opt/cellranger-atac-1.0.1/cellranger-atac", "count"]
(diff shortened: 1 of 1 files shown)


chrisamiller:
Seems straightfoward and looks good!  +1

definitions/tools/cellranger_atac_count.cwl
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Run Cell Ranger ATAC Count"
+
+baseCommand: ["/opt/cellranger-atac-1.0.1/cellranger-atac", "count", "--localmem=64", "--localcores=8"]
```
tmooney: We could possibly set the `--localmem` and `--localcores` based on `runtime.cores` and `runtime.ram`; in some environments the request and how much is given can differ depending on available machine sizes.

apaul7: for converting the `runtime.ram` in mebibytes to gigabytes should I just get a close estimate by dividing by 953.7? or do something more exact like
```
runtime.ram * 1024^2 / 1000^3
```
and in general would it be good to allow a little overhead of requested ram to ram passed in as input to programs or trust them not to go over? if so divide by 960?

tmooney: It would certainly be good to leave a little bit of overhead. Something simple that leaves some room should be fine.

apaul7: done! I'll make another PR to do the same thing for the other cellranger tools

PR #628, opened 2019-04-18
moving cellranger pipeline to pull ram/cpu from runtime env

apaul7:
I still have the Cromwell workflow running, just taking a long time for the test data to land jobs on the cluster. The command generated by Cromwell is the same as it was before. Will wait for the job to finish and succeed before merging.


apaul7:
The `ramMin` resource requirement is in mebibytes. So to convert that to gigabytes you can convert it to bytes it by multiplying by `1024^2`. Then convert the bytes to gigabytes by dividing by `1000^3`. Or you can just divide by `953.6`. I upped it to `1000` to get the extra overhead. 

The original 64000 mebibytes is 67.1gb which gives 3gb overhead.

chrisamiller:
TIL!  +1

PR #632, opened 2019-04-23
germline exome wrapper with hla typing added

johnegarza:


definitions/pipelines/germline_exome_hla_typing.cwl
```diff
@@ -0,0 +1,174 @@
...
+        type: File
+        outputSource: germline_exome_wf/vep_summary
+    optitype_tsv:
+        type: File
+        outputSource: optitype/optitype_tsv
+    optitype_plot:
+        type: File
+        outputSource: optitype/optitype_plot
+steps:
+    germline_exome_wf:
```
tmooney: Echoing [a comment](https://github.com/genome/analysis-workflows/pull/627#discussion_r276324863) comment I left on #627; I generally think we should leave off tags like `wf` from the names of steps.

johnegarza: Fixed in 3709c4a

susannasiebert:
You will also need to update all of the outputs to use the new step name.

PR #633, opened 2019-04-23
Fix some missed connection in workflows and add more secondaryFiles.

tmooney:
Specifying an input that does not exist is merely a warning.  Since we have a lot of optional inputs, this means no error is generated when things are mis-wired.

We should work to eliminate warnings from our CWL validation--we can then possibly make Travis CI fail if new warnings are generated.

### DIFF
--- a/definitions/pipelines/germline_exome.cwl
+++ b/definitions/pipelines/germline_exome.cwl
@@ -148,8 +148,8 @@ steps:
             summary_intervals: summary_intervals
             omni_vcf: omni_vcf
             picard_metric_accumulation_level: picard_metric_accumulation_level   
-            minimum_mapping_quality: qc_minimum_mapping_quality
-            minimum_base_quality: qc_minimum_base_quality
+            qc_minimum_mapping_quality: qc_minimum_mapping_quality
+            qc_minimum_base_quality: qc_minimum_base_quality
         out:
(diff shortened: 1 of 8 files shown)


definitions/pipelines/germline_exome.cwl
```diff
@@ -186,7 +186,7 @@ steps:
             contamination_fraction: extract_freemix/freemix_score
             vep_cache_dir: vep_cache_dir
             synonyms_file: synonyms_file
-            annotate_coding_only: annotate_coding_only
+            coding_only: annotate_coding_only
```
chrisamiller: As of a few versions ago, DV takes `annotate_coding_only` as the param, not just `coding_only`
https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/detect_variants.cwl#L72

Your above detect_variants.cwl file has it as well - old branch?

tmooney: That input goes to `germline_detect_variants.cwl`:
https://github.com/genome/analysis-workflows/blob/afdce1af762f593ad21bc4036483c792ffbc1559/definitions/subworkflows/germline_detect_variants.cwl#L32-L33

chrisamiller: ah, my mistake!   Maybe we should harmonize those params?  (I prefer `annotate_coding_only` as being more explicit about what it actually controls)

tmooney: Sounds good to me--changed this input name everywhere (except the VEP tool itself).

chrisamiller:
+1

tmooney:
Rebased to resolve merge conflict.  Should be pretty much the same as before 😄

PR #637, opened 2019-04-24
Added a new concordance pipeline SOMALIER to compare tumor and normal samples

saimukund20:


definitions/pipelines/somatic_exome.cwl
```diff
@@ -137,6 +137,16 @@ inputs:
         default: true
     manta_output_contigs:
         type: boolean?
+    vcf:
```
tmooney: This is too generic of a name for an input to add to a top-level pipeline.  There are many steps in this workflow and other VCFs are already inputs. It should be made clear what the VCF is.

definitions/tools/concordance.cwl
```diff
@@ -1,48 +1,64 @@
...
+      dockerPull: "brentp/somalier"
+arguments: ["-o", "concordance"]
 inputs:
-    bam_1:
+    vcf:
         type: File
         inputBinding:
+            prefix: "-s"
+            position: 1
+    threads:
```
tmooney: It could be problematic to allow this parameter to have a higher value than the number of cores specified in the `ResourceRequirement`.

definitions/tools/concordance.cwl
```diff
@@ -1,48 +1,64 @@
+#!/usr/bin/env cwl-runner
+
 cwlVersion: v1.0
 class: CommandLineTool
-label: "Concordance tool"
-baseCommand: ["python", "/opt/concordance/newConcordance.py"]
+label: "Concordance checking between Tumor and Normal BAM"
+baseCommand: ["somalier"]
```
tmooney: It might be nice to use a fully-qualified path here.

definitions/tools/concordance.cwl
saimukund20: So then should i remove this input all together?

definitions/pipelines/somatic_exome.cwl
```diff
@@ -331,7 +341,12 @@ outputs:
         type: File?
         outputSource: manta/tumor_only_variants
         secondaryFiles: [.tbi]
-
+    somalier_pairs:
```
chrisamiller: I'd suggest labelling these two outputs something more descriptive that describes what they contain: "concordance_metrics_something..."?

definitions/pipelines/somatic_exome.cwl
```diff
@@ -381,6 +396,19 @@ steps:
...
         out:
             [bam, mark_duplicates_metrics, insert_size_metrics, alignment_summary_metrics, hs_metrics, per_target_coverage_metrics, per_target_hs_metrics, per_base_coverage_metrics, per_base_hs_metrics, summary_hs_metrics, flagstats, verify_bam_id_metrics, verify_bam_id_depth]
+    concordance:
+        run: ../tools/concordance.cwl
+        in:
+            reference: reference
+            bam_1: tumor_alignment_and_qc/bam
+            bam_2: normal_alignment_and_qc/bam
+            vcf: vcf
+            threads: threads
```
chrisamiller: threads probably doesn't need to be a parameter - just set it to something sensible.

definitions/pipelines/somatic_exome.cwl
```diff
@@ -381,6 +396,19 @@ steps:
...
+            min_depth: min_depth
```
chrisamiller: same with some of these other params - probably should just take the two bams and the VCF, set sensible defaults for the rest.

example_data/somatic_exome.yaml
```diff
@@ -75,3 +75,4 @@ interval_list:
   path: exome_workflow/chr17_test_genes.interval_list
 mutect_scatter_count: 2
 mutect_artifact_detection_mode: false
+vcf: "/gscmnt/gc2708/info/medseq/ppg/saimukund/MGI/analysis-workflows/example_data/chr17_vcf_file_for_somalier.vcf"
```
chrisamiller: This should be changed to a relative path "exome_workflow/..."

definitions/tools/concordance.cwl
saimukund20: its a binary. so if we just use the docker container shouldn't it run properly?

tmooney: Theoretically, yes it should! When running Docker there is the option to override environment variables, so it cannot be guaranteed that the `PATH` will be set appropriately.  Some tools require the `PATH` to be corrected. If this one is self-contained, then pointing to the full path inside the container will help eliminate some potential errors.  Adding it isn't required; it just might be nice 😄

saimukund20: How do I find out the path to that binary?

example_data/somatic_exome.yaml
```diff
@@ -75,3 +75,4 @@ interval_list:
   path: exome_workflow/chr17_test_genes.interval_list
 mutect_scatter_count: 2
 mutect_artifact_detection_mode: false
+somalier_vcf: "exome_workflow/chr17_vcf_file_for_somalier.vcf"
```
tmooney: `chr17_vcf_file_for_somalier.vcf` should be added to the `exome_workflow.tgz` file instead of being added directly to the repository if this is the chosen path.

saimukund20: Yes i did it and added the new exome_workflow.tgz

definitions/tools/concordance.cwl
tmooney: You can go inside the docker container to find out.  For instance:
```
$ LSF_DOCKER_PRESERVE_ENVIRONMENT=false docker-interactive brentp/somalier
Job <7938596> is submitted to queue <docker-interactive>.
<<Waiting for dispatch ...>>
<<Starting on blade18-1-15.gsc.wustl.edu>>
Using default tag: latest
latest: Pulling from brentp/somalier
Digest: sha256:eefd6bd95826ce287a1d9201cfa01feaa6982213077c174caa8fba6546daf43f
Status: Image is up to date for brentp/somalier:latest
/usr/bin/id: cannot find name for user ID 12347
/usr/bin/id: cannot find name for group ID 10001
/usr/bin/id: cannot find name for user ID 12347
bash-4.1$ which somalier
~/usr/bin/somalier
bash-4.1$ ls -l /usr/bin/somalier 
-rwxr-xr-x 1 root root 349690 Apr 19 20:32 /usr/bin/somalier
```
so here it's at `/usr/bin/somalier`.

example_data/somatic_exome.yaml
tmooney: I don't see the new one in this PR.  Which commit adds the new version?

definitions/tools/concordance.cwl
```diff
@@ -1,48 +1,44 @@
...
 requirements:
-    - class: DockerRequirement
-      dockerPull: "mgibio/concordance"
     - class: ResourceRequirement
-      ramMin: 4000
+      coresMin: 1
+      ramMin: 8000
+      tmpdirMin: 10000
+    - class: DockerRequirement
+      dockerPull: "brentp/somalier"
```
tmooney: Is there a versioned docker image available?  Relying on `latest` can lead to trouble later.

definitions/tools/concordance.cwl
saimukund20: I did the changed

definitions/tools/concordance.cwl
saimukund20: Added the tag **dockerPull: "brentp/somalier:v0.1.5"** and did the push

tmooney:
The new outputs added to the gathered version of the CWL should also be passed as inputs to the gather step to ensure they are also collected.

definitions/pipelines/gathered_somatic_exome.cwl
```diff
@@ -129,6 +129,8 @@ inputs:
         secondaryFiles: [.tbi]
     output_dir: 
         type: string
+    somalier_vcf:
```
chrisamiller: maybe "concordance_vcf", since that's more generally descriptive than somalier is?

chrisamiller:
Can we get a second review/+1 on this PR, please?  Thanks!

definitions/pipelines/gathered_somatic_exome.cwl
dufeiyu: @chrisamiller  Do you have the real input path for this so we can test it ?

definitions/tools/concordance.cwl
```diff
@@ -1,48 +1,44 @@
...
+        secondaryFiles: [.bai]
 outputs:
-    output_file:
+    somalier_pairs:
         type: File
         outputBinding:
-           glob: "$(inputs.output_file_name)"
-    output_geno_file:
-        type: File?
+            glob: "concordance.somalier.pairs.tsv"
```
dufeiyu: Just a note here and probably out of the scope of this PR. It would be better to add some specific identifier prefix to this hardcoded output name.

definitions/pipelines/gathered_somatic_exome.cwl
chrisamiller: The test data one is in the new gzipped test data file, which is part of this PR.  The real one has been added here:
/gscmnt/gc2560/core/annotation_data/concordance_snps/GRC-human-build38_gnomad_exome_common_snps.vcf

definitions/tools/concordance.cwl
chrisamiller: Yeah, none of this somatic pipeline includes sample names in file names, IIRC.  I'll say that's an issue for another PR.

PR #639, opened 2019-04-30
Add annotation and more filtering to sv pipeline

apaul7:
This PR adds a few things

1) Adds VEP annotation to the merged SV vcf. 
	Specifically only interested in the gene symbols. Which required putting the `--everything` flag back as an optional input and defaulting to `True`, reverting #552.
2) Allow input of an array of interval list for filtering which generates a named vcf using the special labelled_file type
3) Adds vcf to tsv for those interval list vcfs
4) Added an optional prefix for the resulting tsv in `tools/add_vep_fields_to_table.cwl` to avoid name collisions in results
5) Wired the `pipelines/germline_wgs.cwl` with new inputs/outputs

Have tested this on a small down sampled bam and currently running with a full sized one.
Tried to keep most things as defaults in the sv workflow to reduce the number of new inputs required to run the updated cwl.


apaul7:
Added a sorting step for the merged sv vcf. Which allowed the wgs sample to complete `subworkflows/single_sample_sv_callers.cwl` successfully.

Also updated the survivor docker image.

One other thing to note is that I altered the `tools/sort_vcf.cwl` to make the reference sequence dictionary an optional input instead of requiring it as an input in the whole workflow.

PR #642, opened 2019-05-03
add tsv step to germline detect variants

apaul7:
This adds the tsv file generation for the germline detect variants


apaul7:
I had to update the pipeline cwls to pass the table field values to the subworkflows. I had them as inputs but did not have them connected

PR #643, opened 2019-05-03
New tool to add arbitrary text to a file at a specific line

susannasiebert:
Please note: this tool won't work on gzipped files so for the VCF I'll have to dig into whether these are gzipped at the point in the workflow that we want to add the CLE disclaimer and add a second version that first `gzcat`s the input_file and then pipes the output to awk.


susannasiebert:
I just pushed a second version of this tool which will take a bgzipped file as an input. @apaul7 can you re-review please?

PR #649, opened 2019-05-07
rnaseq workflow with star aligner

sridhar0605:
Add rna-seq workflow with STAR aligner.

- [x]  Add appropriate flag for downstream Stringtie Quantification

- [x]  Add appropriate flag for downstream STAR-Fusion analysis 

**Action:**
1. Added `outSAMstrandField: intronMotif` that adds `XS` tags to each read required by Stringtie

2. This is a bit tricky currently STAR fusion does not accept aligned bam, however it accepts R1 and R2 
`--readFilesIn ${left_fq_filename} ${right_fq_filename}`;  how should we tackle this for reads from different lanes? 
Option A) easier : https://github.com/STAR-Fusion/STAR-Fusion/issues/125

Option B) This get complicated if we try and add RG's for each lane STAR current format is 
`--outSAMattrRGline:
ID:sampleA CN:AA DS:AAA , ID:sampleBB CN:bb DS:bbbb , ID:sampleC CN:ccc DS:cccc `
commas have to be separated by spaces on left and right.


definitions/pipelines/rnaseq_star.cwl
```diff
@@ -0,0 +1,150 @@
...
+    outSAMattrRGline:
+        type:
+            type: array
+            items:
+                type: array
+                items: string
+    genomeDir:
+        type: Directory
+    sjdbGTFfile:
+        type: File
```
chrisamiller: This isn't a very informative name.  Could you make this a little more human readable?  Is it just the gtf?  If so, maybe `gtf_file` or `transcriptome_gtf_file` or something

definitions/pipelines/rnaseq_star.cwl
```diff
@@ -0,0 +1,150 @@
...
+    sample_name:
+        type: string
+    runMode:
+        type: string
```
chrisamiller: Is this ever going to be anything other than `alignReads` for this workflow?  If not then we should drop this at the top level and make the inputs as simple as we can.

definitions/pipelines/rnaseq_star.cwl
```diff
@@ -0,0 +1,150 @@
...
+    outFileNamePrefix:
+        type: string
```
chrisamiller: this could probably be hard-coded in the subtool, or at least have a default

definitions/pipelines/rnaseq_star.cwl
```diff
@@ -0,0 +1,150 @@
...
+    outSAMstrandField:
+        type: string
```
chrisamiller: any reason not to set a default here, or to hard code it?

definitions/pipelines/rnaseq_star.cwl
```diff
@@ -0,0 +1,150 @@
...
+    instrument_data_bams:
+        type: File[]
+    outSAMattrRGline:
+        type:
+            type: array
+            items:
+                type: array
+                items: string
+    genomeDir:
+        type: Directory
```
chrisamiller: is this the directory path to the reference? If so, make that clear in the variable name

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,208 @@
...
+    log:
+        type: File
+        outputBinding:
+          glob: "$(inputs.outFileNamePrefix)Log.out"
+    log_progress:
+        type: File
+        outputBinding:
+          glob: "$(inputs.outFileNamePrefix)Log.progress.out"
+    sj:
+        type: File
```
chrisamiller: this name could be more informative

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,208 @@
...
+        outputBinding:
+          glob: "$(inputs.outFileNamePrefix)Log.final.out"
+    log:
+        type: File
+        outputBinding:
+          glob: "$(inputs.outFileNamePrefix)Log.out"
+    log_progress:
+        type: File
+        outputBinding:
+          glob: "$(inputs.outFileNamePrefix)Log.progress.out"
```
chrisamiller: what are these logs? Are they useful above and beyond what gets output to stdout/stderr and captured by cromwell?  Is there a reason to expose them as an output (as opposed to just leaving them in the execution directory for debugging?)

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,208 @@
...
+label: "STAR: align reads to transcriptome"
+baseCommand: STAR
+requirements:
+    - class: ShellCommandRequirement
+    - class: ResourceRequirement
+      ramMin: 32000
+      coresMin: 16
+    - class: DockerRequirement
+      dockerPull: "sridnona/mgirnaseq:v1"
+arguments: [
```
chrisamiller: I'll suggest that we push this to mgibio if it's a dockerfile you've created. Just helpful to keep core pipeline dockers all in one place!

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,208 @@
...
+class: CommandLineTool
+cwlVersion: v1.0
+label: "STAR: align reads to transcriptome"
+baseCommand: STAR
+requirements:
+    - class: ShellCommandRequirement
+    - class: ResourceRequirement
+      ramMin: 32000
+      coresMin: 16
+    - class: DockerRequirement
```
chrisamiller: I'm surprised that this only takes 2G per core - have you run it enough to know that'll hold for most samples?

definitions/pipelines/rnaseq_star.cwl
sridhar0605: made it default

definitions/tools/star_align.cwl
sridhar0605: changed it to splicejunction out

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,208 @@
...
+           None        ... 1-pass mapping
+           Basic ... basic 2-pass mapping, with all 1st pass junctions inserted into the genome indices on the fly
+    outFilterMultimapNmax:
+        type: string?
+        inputBinding:
+            position: 13
+            prefix: '--outFilterMultimapNmax'
+        doc: '
+
+      int: read alignments will be output only if the read maps fewer than this value,
```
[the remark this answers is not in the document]
sridhar0605: good catch changed it to `int?`

definitions/tools/star_align.cwl
sridhar0605: IIRC these logs are good files that can be fed to multiqc.

definitions/pipelines/rnaseq_star.cwl
sridhar0605: changed it to stargenomeDir

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,190 @@
...
+            position: 18
+            prefix: '--outSAMattributes'
+        doc: |
+          Standard
+          string: a string of desired SAM attributes, in the order desired for the output SAM
+          NH HI AS nM NM MD jM jI XS ... any combination in any order
+          Standard   ... NH HI AS nM
+          All        ... NH HI AS nM NM MD jM jI
+          None ... no attributes
+    outSAMattrRGline:
```
johnegarza: Looking at the sample yaml you updated, it looks like this is intended to take in an array of arrays of strings, but this doesn't match up with the type here. Additionally, according to the documentation, each RG line should be separated by " , " which isn't the current behavior.  CWL has an itemSeparator field that will allow us to insert " , " between array elements, but I can't get that working with nested arrays in my own tests. Maybe this should just be an array of strings, where each string is the full RG line- in which case the yaml should be updated.

example_data/rnaseq_star/workflow.yaml
```diff
@@ -0,0 +1,46 @@
+instrument_data_bams:
+    - class: File
+      path: test.bam
+    - class: File
+      path: test2.bam
+outSAMattrRGline: 
+    - ['ID:2895626107', 'PU:H3MYFBBXX5.4.GCCAAT', 'SM:H_NJ-HCC1395-HCC1395_RNA', 'LB:Pooled_RNA_2891007020-cD1-lib1', 'PL:Illumina', 'CN:WUGSC']
```
johnegarza:
```suggestion
    - ['ID:2895626107 PU:H3MYFBBXX5.4.GCCAAT SM:H_NJ-HCC1395-HCC1395_RNA LB:Pooled_RNA_2891007020-cD1-lib1 PL:Illumina CN:WUGSC']
```
See the comment about the outSAMattrRGline field in tools/star_align.cwl to see why I suggested this

definitions/tools/star_align.cwl
```diff
@@ -0,0 +1,190 @@
...
+        doc: 'int>0: length of the donor/acceptor sequence on each side of the junctions, ideally = (mate_length - 1)
+        '
+    gtf_file:
+        type: File?
+        inputBinding:
+            position: 15
+            prefix: '--sjdbGTFfile'
+        doc: |
+          string: path to the GTF file with annotations
+    outFileNamePrefix:
```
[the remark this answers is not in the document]
sridhar0605: this not a default for star because you would have supplied the GTF during index creation step. 
However for Stringtie step this is default. But I see the point just to keep things consistent ill make this default.

chrisamiller:
Is it safe to close this PR, @sridhar0605 ?

sridhar0605:
Yes ill close it

PR #653, opened 2019-05-08
strips trailing empty lines that may cause problems

chrisamiller:
If there's a stray extra newline at the end of an interval list, that gets put into the chromosomes list and causes steps like Pindel to choke (when given an empty string for the -c argument)

The underlying fix is really "fix the interval list", but there's no reason not to make this a little more robust as well


chrisamiller:
No issue - just me trying to get the mouse somatic exome set up and doing some testing

PR #655, opened 2019-05-09
moving accessory script to CWL for transparency

chrisamiller:


chrisamiller:
If anytime we have to go dig up the source of one of these scripts, we just drop it in, we'll get these all cleaned up eventually!

PR #661, opened 2019-05-16
Upgrade to version ~1.4.0~ 1.4.1

susannasiebert:


definitions/tools/pvacvector.cwl
```diff
@@ -84,6 +84,16 @@ inputs:
         inputBinding:
             prefix: "--n-threads"
         default: 8
+    spacers:
+        type: string?
+        inputBinding:
+            prefix: "--spacers"
+        default: "None,AAY,HHHH,GGS,GPGPG,HHAA,AAL,HH,HHC,HHH,HHHD,HHL,HHHC"
```
tmooney: Would it make sense to make this a `string[]` with `itemSeparator: ","` instead?

chrisamiller:
+1

PR #662, opened 2019-05-16
New immuno workflow

johnegarza:
This is the first attempt at making a new immuno workflow, combining 5 of our other workflows. There is some overlap/redundancy between a few of these, but for now we'd like to get a functioning workflow. Reviews focusing on checking all of the inputs would be appreciated, see comments within immuno.cwl, especially in the pvacseq step at the bottom. A test case should also be developed soon.

### DIFF
--- a/definitions/pipelines/immuno.cwl
+++ b/definitions/pipelines/immuno.cwl
@@ -42,7 +42,10 @@ inputs:
     gene_transcript_lookup_table:
        type: File
     strand:
-       type: string?
+        type:
+          - "null"
+          - type: enum
+            symbols: ["first", "second", "unstranded"]
     refFlat:
         type: File
     ribosomal_intervals:
--- a/definitions/pipelines/immuno.cwl
+++ b/definitions/pipelines/immuno.cwl
@@ -4,6 +4,9 @@ cwlVersion: v1.0
 class: Workflow
 label: "Immunotherapy Workflow"
 requirements:
+    - class: SchemaDefRequirement
+      types:
+          - $import: ../types/labelled_file.yml
     - class: SubworkflowFeatureRequirement
 inputs:
     reference_index:
@@ -412,6 +415,12 @@ steps:
             bam: somatic/tumor_cram
         out:
             [phased_vcf]
+    extract_alleles:
+        run: ../tools/extract_hla_alleles.cwl
+        in:
+            allele_file: germline/optitype_tsv
+        out:
+            [allele_string]
     pvacseq:
         run: ../subworkflows/pvacseq.cwl
         in:
@@ -424,7 +433,7 @@ steps:
             readcount_minimum_mapping_quality: readcount_minimum_mapping_quality #should this be different from qc_minimum_mapping_quality
             gene_expression_file: rnaseq/stringtie_gene_expression_tsv #not sure if this connection is correct
             transcript_expression_file: rnaseq/transcript_abundance_tsv #not sure if this connection is correct
-            allele_file: germline/optitype_tsv
+            alleles: extract_alleles/allele_string
             prediction_algorithms: prediction_algorithms
             epitope_lengths: epitope_lengths
             binding_threshold: binding_threshold
--- a/definitions/pipelines/immuno.cwl
+++ b/definitions/pipelines/immuno.cwl
@@ -1,0 +1,283 @@
+#/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "Immunotherapy Workflow"
+requirements:
+    - class: SubworkflowFeatureRequirement
+inputs:
+    reference: string
+    reference_index:
+        type: File #this requires an extra file with the basename
+        secondaryFiles: [".1.ht2", ".2.ht2", ".3.ht2", ".4.ht2", ".5.ht2", ".6.ht2", ".7.ht2", ".8.ht2"]
+    reference_annotation:
+        type: File
+    rna_bams:
+        type: File[]
+    tumor_bams:
+        type: File[]
+    normal_bams:
+        type: File[]
+    rna_readgroups:
+        type: string[]
+    tumor_readgroups:
+        type: string[]
+    normal_readgroups:
+        type: string[]
+    sample_name:
+        type: string
+    trimming_adapters:
+        type: File
+    trimming_adapter_trim_end:
+        type: string
+    trimming_adapter_min_overlap:
+        type: int
+    trimming_max_uncalled:
+        type: int
+    trimming_min_readlength:
+        type: int
+    kallisto_index:
+       type: File
+    gene_transcript_lookup_table:
+       type: File
+    strand:
+       type: string?
+    refFlat:
+        type: File
+    ribosomal_intervals:
+        type: File
+outputs:
+    annotated_vcf:
+        type: File
+        outputSource: pvacseq/annotated_expression_vcf
+    annotated_tsv:
+        type: File
+        outputSource: pvacseq/annotated_variants_tsv
+    mhc_i_all_epitopes:
+        type: File?
+        outputSource: pvacseq/mhc_i_all_epitopes
+    mhc_i_filtered_epitopes:
+        type: File?
+        outputSource: pvacseq/mhc_i_filtered_epitopes
+    mhc_i_ranked_epitopes:
+        type: File?
+        outputSource: pvacseq/mhc_i_ranked_epitopes
+    mhc_ii_all_epitopes:
+        type: File?
+        outputSource: pvacseq/mhc_ii_all_epitopes
+    mhc_ii_filtered_epitopes:
+        type: File?
+        outputSource: pvacseq/mhc_ii_filtered_epitopes
+    mhc_ii_ranked_epitopes:
+        type: File?
+        outputSource: pvacseq/mhc_ii_ranked_epitopes
+    combined_all_epitopes:
+        type: File?
+        outputSource: pvacseq/combined_all_epitopes
+    combined_filtered_epitopes:
+        type: File?
+        outputSource: pvacseq/combined_filtered_epitopes
+    combined_ranked_epitopes:
+        type: File?
+        outputSource: pvacseq/combined_ranked_epitopes
+steps:
+    rnaseq:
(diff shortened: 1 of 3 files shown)


susannasiebert:
Do you have an example of the optitype output file?

johnegarza:
Example optitype output:
```
        A1      A2      B1      B2      C1      C2      Reads   Objective
0       A*26:01 A*02:05 B*55:01 B*50:01 C*03:03 C*06:02 3684.0  3485.0640000000008
```

definitions/pipelines/immuno.cwl
```diff
@@ -93,191 +306,151 @@ steps:
...
+            reference_dict: reference_dict
             bam: somatic/tumor_cram
         out:
             [phased_vcf]
     pvacseq:
-        run: pvacseq.cwl
+        run: ../subworkflows/pvacseq.cwl
         in:
             detect_variants_vcf: somatic/final_vcf
+            sample_name: sample_name #should this be the same as the above sample_name
```
susannasiebert: These needs to be tumor's sample name in the pVACseq input VCF. I believe in our pipelines this ends up being TUMOR.

definitions/pipelines/immuno.cwl
```diff
@@ -93,191 +306,151 @@ steps:
...
+            normal_sample_name: normal_name #is this right?
```
susannasiebert: These needs to be normal sample name in the pVACseq input VCF. I believe in our pipelines this ends up being NORMAL.

definitions/subworkflows/pvacseq.cwl
```diff
@@ -30,8 +31,10 @@ inputs:
     expression_tool:
         type: string?
         default: 'kallisto'
-    alleles:
-        type: string[]
+    allele_file:
```
susannasiebert: I wonder if this parsing needs to be it's own step so that we can preserve the pVACseq CWL with alleles as an input for cases where we need to run this workflow directly (e.g., when we have clinical HLA typing information or want to run class I and class II alleles [which is not supported by optitype]).

johnegarza: Changed in latest commit

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,467 @@
...
+                default: 'TUMOR'
+            normal_sample_name: 
+                default: 'NORMAL'
+            rnaseq_bam: rnaseq/final_bam
+            reference_fasta: reference
+            readcount_minimum_base_quality: readcount_minimum_base_quality #should this be different from qc_minimum_base_quality?
+            readcount_minimum_mapping_quality: readcount_minimum_mapping_quality #should this be different from qc_minimum_mapping_quality
+            gene_expression_file: rnaseq/stringtie_gene_expression_tsv #not sure if this connection is correct
+            transcript_expression_file: rnaseq/transcript_abundance_tsv #not sure if this connection is correct
+            alleles: extract_alleles/allele_string
```
chrisamiller: Since we often do clinical HLA-typing, is there a reasonable way to allow the user to provide an HLA string, that overrides whatever Optitype finds?

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,467 @@
...
+            - type: enum
+              symbols: ["cterm", "20s"]
+    net_chop_threshold:
+        type: float?
+    netmhc_stab:
+        type: boolean?
+    pvacseq_threads:
+        type: int?
+
+outputs:
```
chrisamiller: It seems like this leaves out a lot of outputs that would be useful.  (bam files, qc stats, expression outputs, etc).  If we don't save those as outputs, they all get trashed, and that's not what we want. 

 For example, what if we need to look in the bams to see if the reads support a variant that produces good epitope?  Or need to diagnose why we got very few epitopes?  

My feeling is that really, we should be passing through just about every output from the sub-workflows into the final directory.  It'd be better yet if we can come up with a reasonable sub-directory organization for them, rather than having one outputs directory with a hundred files.

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,467 @@
...
+
+cwlVersion: v1.0
+class: Workflow
+label: "Immunotherapy Workflow"
+requirements:
+    - class: SchemaDefRequirement
+      types:
+          - $import: ../types/labelled_file.yml
+    - class: SubworkflowFeatureRequirement
+inputs:
```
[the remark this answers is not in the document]
johnegarza: Added

definitions/pipelines/immuno.cwl
johnegarza: Added them all to final outputs

johnegarza:
@susannasiebert or anyone else familiar with these workflows- I hardcoded in the sample values as 'TUMOR' and 'NORMAL'. Is this safe for now or can we make them dynamic? Do the names come from the gms, or [these lines in the somatic workflow?](https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/detect_variants.cwl#L319)

definitions/pipelines/immuno.cwl
johnegarza: Since there are still no conditionals in CWL, I'm not sure if there is a reasonably simple way to do this. Suggestions from anyone? Maybe provide `auto_alleles` and `manual_alleles` as inputs to the tool, where manual is optional and overrides auto if present?

susannasiebert:
@johnegarza the line you referenced doesn't set the TUMOR and NORMAL name, it just references what's already in the VCF header. I'm not sure which step precisely sets the sample names in the VCF header. It might be the variant callers themselves? @jasonwalker80 do you know?

jasonwalker80:
Somatic variant callers default to `NORMAL` and `TUMOR` sample names/identifiers. We will want to use our GMS sample identifiers at some point so we can combine VCFs from DNA/RNA, multiple timepoints, etc.

definitions/pipelines/immuno.cwl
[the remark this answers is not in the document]
tmooney: If we wanted, we could use records to group these.  Like:
```yaml
rna:
  type: record
  name: rna
  fields:
    reference_index:
      type: File
      #etc.
    reference_annotation:
      type: File
    #etc.
somatic:
  type: record
  name: somatic
  fields:
    reference:
      type: string
      #etc.
  #etc.
#etc.
```
Then the inputs YAML would also be structured.
```yaml
rna:
  reference_index:
    #etc.
  reference_annotation:
    #etc.
  #etc.
somatic:
  reference: "something"
  #etc.
#etc.
```

If we decide to go this route and start reusing these records in multiple places, we'll likely need to convert them to types under <https://github.com/genome/analysis-workflows/tree/master/definitions/types>.

definitions/pipelines/immuno.cwl
johnegarza: Following discussion, this will be added in v2

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,777 @@
...
+        type:
+            - "null"
+            - type: enum
+              symbols: ["cterm", "20s"]
+    net_chop_threshold:
+        type: float?
+    netmhc_stab:
+        type: boolean?
+    pvacseq_threads:
+        type: int?
```
chrisamiller: Is there a reason to expose this?  In general, our practice has seemed to be to set appropriate resource requests in the CWL for number of threads and memory.

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,777 @@
...
+            sample_name: 
+                default: 'TUMOR'
+            normal_sample_name: 
+                default: 'NORMAL'
+            rnaseq_bam: rnaseq/final_bam
+            reference_fasta: reference
+            readcount_minimum_base_quality: readcount_minimum_base_quality #should this be different from qc_minimum_base_quality?
+            readcount_minimum_mapping_quality: readcount_minimum_mapping_quality #should this be different from qc_minimum_mapping_quality
+            gene_expression_file: rnaseq/stringtie_gene_expression_tsv #not sure if this connection is correct
+            transcript_expression_file: rnaseq/transcript_abundance_tsv #not sure if this connection is correct
```
chrisamiller: These connections are commented as "not sure".  If you are now sure and this has been tested, it seems like we should remove these comments!

susannasiebert: These look like the correct connections to me.

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,777 @@
...
+        run: ../subworkflows/pvacseq.cwl
+        in:
+            detect_variants_vcf: somatic/final_vcf
+            sample_name: 
+                default: 'TUMOR'
+            normal_sample_name: 
+                default: 'NORMAL'
+            rnaseq_bam: rnaseq/final_bam
+            reference_fasta: reference
+            readcount_minimum_base_quality: readcount_minimum_base_quality #should this be different from qc_minimum_base_quality?
```
susannasiebert: the qc quality thresholds are for a different tool. So I would keep the parameters separate at least for now.

definitions/pipelines/immuno.cwl
```diff
@@ -0,0 +1,777 @@
...
+            trna_cov: trna_cov
+            normal_vaf: normal_vaf
+            tdna_vaf: tdna_vaf
+            trna_vaf: trna_vaf
+            expn_val: expn_val
+            net_chop_method: net_chop_method
+            net_chop_threshold: net_chop_threshold
+            netmhc_stab: netmhc_stab
+            n_threads: pvacseq_threads
+            variants_to_table_fields: variants_to_table_fields #the defaults set in this file will override the defaults for this parameter set in the pvacseq subworkflow; is this ok?
```
susannasiebert: I think we'd want to use consistent output fields between the different variants to table steps so overwriting the defaults is ok.

definitions/pipelines/immuno.cwl
susannasiebert: We're still trying to figure out what those numbers should be. Until we do I think it's ok to expose this to give us the flexibility to easily change this parameter.

PR #665, opened 2019-05-20
Added the makrdup step in the RNAseq pipeline

saimukund20:


definitions/pipelines/rnaseq.cwl
```diff
@@ -52,8 +52,11 @@ inputs:
 outputs:
     final_bam:
         type: File
-        outputSource: index_bam/indexed_bam
+        outputSource: mark_dup/sorted_bam
         secondaryFiles: [.bai]
+    markdup_metrics:
```
chrisamiller: I guess it doesn't hurt to output these metrics, but it's probably not really needed!

definitions/pipelines/rnaseq.cwl
```diff
@@ -129,10 +132,16 @@ steps:
             bam: merge/merged_bam
         out:
             [indexed_bam]
+    mark_dup:
```
chrisamiller: We don't need to output a separate markdup bam - just need to make sure the final bam is derived from it (as you did above!)

definitions/pipelines/rnaseq.cwl
```diff
@@ -129,10 +132,16 @@ steps:
...
+        run: ../tools/mark_duplicates_and_sort.cwl
+        in:
+            bam: merge/merged_bam
+        out:
+            [sorted_bam, metrics_file]
     stringtie:
         run: ../tools/stringtie.cwl
         in:
-            bam: merge/merged_bam
+            bam: mark_dup/sorted_bam
```
chrisamiller: Did you verify that stringtie ignores duplicate marking?  (either in their docs/online or by testing it yourself?)

definitions/pipelines/rnaseq.cwl
chrisamiller: my bad - I misread this. it's fine!

definitions/pipelines/rnaseq.cwl
saimukund20: removed that as an output.

definitions/pipelines/rnaseq.cwl
saimukund20: I couldn't find any literature of stringtie that says it ignores duplicate marking, but i ran a test and the outputs with/without markdup were the same.

PR #666, opened 2019-05-20
Add three report cwl for aml_trio

dufeiyu:
These three report cwl are needed for upcoming aml_trio cwl pipeline.


dufeiyu:
It makes sense. Also Alex points out that input File[] in cwl sometimes could be different order. I will make some changes to better error checking and check output file identity in the perl script and then update the docker image tag.

dufeiyu:
Yes, v1.0 now has the updated perl scripts.

PR #671, opened 2019-05-24
updating fp_filter to version 0.1.1 with bugfix

chrisamiller:


apaul7:
shouldn't the docker image be `1.0.1`? instead of downgrading to `0.1.1` from `1.0.0`

chrisamiller:
The fp_filter tool was listed as 0.1.0, so the bugfix took it to 0.1.1, and then I was trying to enforce some consistency between that and the repo containing the docker wrapper for that tool.  I suppose the apparent downgrade is confusing - maybe it's better to update both repos to 1.0.1, and then make them match

chrisamiller:
Okay, updated the relevant tools to be 1.0.1 for clarity.  New commit fixes that here as well.

PR #673, opened 2019-05-24
cleaning up dirs, adding process_input.pl scripts

chrisamiller:
This PR does two things:
1) removes the confusing (and largely useless, as far as I could tell) workflow directories from the main folder
2) adds a directory that contains process_inputs.pl files for a few of the main workflows.  This a) gets them in version control and b) makes them easy to update along with PRs to the corresponding CWLs.


gms_process_inputs/somatic_exome.process_inputs.pl
```diff
@@ -0,0 +1,184 @@
...
+}
+my $fl;
+
+if ($roi_input) {
+    $fl = Genome::FeatureList->get(name => $roi_input->value_id)
+        or $build->fatal_message('no ROI found for input');
+} elsif ($multiple_trsn) {
+    $build->fatal_message('Multiple target_region_set_name found on instrument data. Please specify a "region_of_interest_set_name" input explicitly.');
+} elsif ($target_region_set_name) {
+    $fl = Genome::FeatureList->get(name => $target_region_set_name)
```
tmooney: This needs to handle newer LIMS target sets where the ID is tossed in the instrument data's `target_region_set_name` column.
```suggestion
    $fl = Genome::FeatureList->get(name => $target_region_set_name) // Genome::FeatureList->get(id => $target_region_set_name)
```

gms_process_inputs/rnaseq.process_inputs.pl
```diff
@@ -0,0 +1,82 @@
...
+my $build_id = $ARGV[0]
+    or die 'no build id';
+
+my $build = Genome::Model::Build->get($build_id)
+    or die 'no build for id';
+
+my @inputs = $build->inputs;
+
+my %inputs;
+for my $file_input ('reference_index', 'trimming_adapters', 'reference_annotation', 'kallisto_index', 'gene_transcript_lookup_table','refFlat','ribosomal_intervals') {
```
tmooney: It'd be better if the inputs here used the `InputProcessor` strategy where possible like in the `somatic_exome.process_inputs.pl` in this PR.

chrisamiller: Since this is the latest version of the process script currently in the pipeline, I think I'm going to leave it here and punt on doing this for now. Next time we update the PP, I completely agree that we should change it to use that strategy.

PR #676, opened 2019-05-28
Support multiple species and builds with VEP

chrisamiller:
This is a big one, since it touches anything that uses VEP, which is most workflows. 

We previously assumed that we'd point our tool to a VEP cache, and that it would always be human, and always match the version referred to in the vep tool docker version.  This doesn't work for mouse, or other species, and limits our ability to support multiple projects that may want to anchor themselves to a particular VEP build.

The solution is to expose a bunch of parameters for reference assembly, ensembl version, and species, so that they must be explicitly specified.  While I don't love adding extra complexity to the (already kind of overwhelming) input yamls, asking users to specify the  ensembl version and such seems like a reasonable ask, IMO.

### DIFF
--- a/definitions/pipelines/cle_somatic_exome.cwl
+++ b/definitions/pipelines/cle_somatic_exome.cwl
@@ -104,13 +104,13 @@ inputs:
         type: string
     vep_ensembl_assembly:
         type: string
-        doc: genome assembly to use in vep. Examples: "GRCh38" or "GRCm38"
+        doc: "genome assembly to use in vep. Examples: GRCh38 or GRCm38"
     vep_ensembl_version:
         type: string
-        doc: ensembl version - Must be present in the cache directory. Example: "95"
+        doc: "ensembl version - Must be present in the cache directory. Example: 95"
     vep_ensembl_species:
         type: string
-        doc: ensembl species - Must be present in the cache directory. Examples: "homo_sapiens" or "mus_musculus"
+        doc: "ensembl species - Must be present in the cache directory. Examples: homo_sapiens or mus_musculus"
     synonyms_file:
         type: File?
     annotate_coding_only:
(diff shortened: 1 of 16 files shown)


chrisamiller:
> Is the version really something we should parameterize when it's hardcoded in the Docker image: 
Good thoughts. John and I had a conversation about this in slack:
https://mgibio.slack.com/archives/C1AHCQM5X/p1558713693016200

The gist of it is, that we essentially have a choice between maintaining separate VEP caches, and having many different VEP CWLs pointing to that docker container (or branches, etc),   or having one cache and making it a parameter.  The good news is that it should all be backwards compatible (and in most cases, forward compatible as well).  So even though the vep docker version is 95, it happily runs on mouse 96, for example, and we've used it as far back as versions in the 60s, IIRC.

> This type of change with a collection of related parameters is really making me think a record type (like the "labelled file" might be handy)!
I don't know anything about those, but would be happy to learn! Maybe make an issue for it and we'll look at a future PR?

chrisamiller:
I should add that we also should update that VEP docker version from time to time (probably when we prepare a "major release" of the pipelines, like we're doing around version 95)

tmooney:
A record looks something like this:
https://github.com/genome/analysis-workflows/blob/92bdcd9fa879161834ecdb1c4c9ac7c46e940206/definitions/types/labelled_file.yml#L1-L8
and then can be used with a combination of an import like https://github.com/genome/analysis-workflows/blob/92bdcd9fa879161834ecdb1c4c9ac7c46e940206/definitions/pipelines/somatic_exome.cwl#L7-L9 and a reference like https://github.com/genome/analysis-workflows/blob/92bdcd9fa879161834ecdb1c4c9ac7c46e940206/definitions/pipelines/somatic_exome.cwl#L43-L44 (this example uses an array of them).

chrisamiller:
Added this issue :
https://github.com/genome/analysis-workflows/issues/679

PR #678, opened 2019-05-28
Added ASSUME_SORT_ORDER as an input

saimukund20:
I changed the ASSUME_SORT_ORDER as user defined because the value for ASSUME_SORT_ORDER varies between the rnaseq and somatic_exome pipelines.

### DIFF
--- a/definitions/pipelines/rnaseq.cwl
+++ b/definitions/pipelines/rnaseq.cwl
@@ -40,9 +40,6 @@ inputs:
         type: File
     gene_transcript_lookup_table:
         type: File
-    markdup_assume_sort_order:
-        type: string?
-        default: "coordinate"
     strand:
         type:
           - "null"
@@ -136,7 +133,7 @@ steps:
         run: ../tools/mark_duplicates_and_sort.cwl
         in:
             bam: merge/merged_bam
-            markdup_assume_sort_order: markdup_assume_sort_order
+            input_sort_order: "coordinate"
         out:
             [sorted_bam, metrics_file]
     stringtie:
--- a/definitions/pipelines/rnaseq.cwl
+++ b/definitions/pipelines/rnaseq.cwl
@@ -40,6 +40,9 @@ inputs:
         type: File
     gene_transcript_lookup_table:
         type: File
+    markdup_assume_sort_order:
+        type: string?
+        default: "queryname"
     strand:
         type:
           - "null"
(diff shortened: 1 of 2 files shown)


definitions/tools/mark_duplicates_and_sort.cwl
```diff
@@ -37,6 +37,10 @@ inputs:
         type: File
         inputBinding:
             position: 1
+    markdup_assume_sort_order:
```
chrisamiller: `input_sort_order` might be a bit more understandable

definitions/tools/mark_duplicates_and_sort.cwl
```diff
@@ -37,6 +37,10 @@ inputs:
...
+        type: string
+        inputBinding:
+            position: 5
```
chrisamiller: I'd favor giving this a default value of "queryname", which is what it was previously (and what several pipelines expect).  Then you can explicitly override it in the RNAseq pipeline, which now inputs "coordinate"

definitions/tools/mark_duplicates_and_sort.cwl
```diff
@@ -21,9 +21,9 @@ requirements:
             declare MD_BARCODE_TAG
             if [ ! -z "${5}" ]; then
               MD_BARCODE_TAG="BARCODE_TAG=${5}"
```
chrisamiller: If you're adding the sortorder as a param, the above two need to be changed to `$6`

definitions/pipelines/rnaseq.cwl
```diff
@@ -40,6 +40,9 @@ inputs:
         type: File
     gene_transcript_lookup_table:
         type: File
+    markdup_assume_sort_order:
+        type: string?
+        default: "coordinate"
```
chrisamiller: This doesn't need to be an input, since it's hardwired into the way this pipeline runs.

definitions/pipelines/rnaseq.cwl
```diff
@@ -133,6 +136,7 @@ steps:
         run: ../tools/mark_duplicates_and_sort.cwl
         in:
             bam: merge/merged_bam
+            markdup_assume_sort_order: markdup_assume_sort_order
```
chrisamiller: Just hardcode this to the appropriate value for this pipeline, doesn't need to be an input.

Issue #674, opened 2019-05-24
Reference string in somatic-exome.cwl


Stikus:
Hello.
According to these lines in files: [somatic_exome.cwl#L13](https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/somatic_exome.cwl#L13), [detect_variants.cwl#L10](https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/detect_variants.cwl#L10), [varscan_pre_and_post_processing.cwl#L11](https://github.com/genome/analysis-workflows/blob/master/definitions/subworkflows/varscan_pre_and_post_processing.cwl#L11), [varscan.cwl#L7](https://github.com/genome/analysis-workflows/blob/master/definitions/subworkflows/varscan.cwl#L7) 'reference' is a 'string', but for real it is still a 'file' and [this](https://github.com/genome/analysis-workflows/blob/master/definitions/tools/varscan_somatic.cwl#L6) script calling 'samtools mpileup' with '-f' flag, which specifies 'file', not 'string'.
I assume that 'reference' string from file [somatic_exome.yaml#L59](https://github.com/genome/analysis-workflows/blob/master/example_data/somatic_exome.yaml#L59) is path to some directory that is mounted to docker-container by default or after some docker-engine settings. Is it correct?
I think so because CWL-files dont specify 'reference' as file and cwltool dont transfer this 'reference' to docker-container as file.

If I'm correct - can you tell how to run your workflow without this 'reference' in docker-containers? Do I need to modify all CWL-files to switch 'reference' type from 'string' to 'File'?

And is 'chr17_test.fa' file in [exome_workflow.tgz](https://github.com/genome/analysis-workflows/blob/master/example_data/exome_workflow.tgz) same as  'chr17_test.fa' file from [somatic_exome.yaml#L59](https://github.com/genome/analysis-workflows/blob/master/example_data/somatic_exome.yaml#L59)?

johnegarza:
[...]

PR #682, opened 2019-05-29
Support for both 'File' and 'string' types of 'reference' input in  varscan_pre_and_post_processing.cwl

serge2016:
For better compatibility of the repo with other users in the internet I suggest to add both types (`string` and `File` with `secondaryFiles`) for `reference` input.
For now (in this PR) just to show the possibility one subworkflow was modified: `varscan_pre_and_post_processing.cwl`.

The correctness of this CWL code is discussed: https://github.com/common-workflow-language/common-workflow-language/issues/858.

The purpose of both types for this input is described: https://github.com/genome/analysis-workflows/issues/674.

### DIFF
--- a/definitions/pipelines/exome_alignment.cwl
+++ b/definitions/pipelines/exome_alignment.cwl
@@ -9,7 +9,11 @@ requirements:
           - $import: ../types/labelled_file.yml
     - class: SubworkflowFeatureRequirement
 inputs:
-    reference: string
+    reference:
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai, ^.dict, .amb, .ann, .bwt, .pac, .sa]
     bams:
         type: File[]
     readgroups:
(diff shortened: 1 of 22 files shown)


chrisamiller:
Thanks for the PR, serge2016!  This is interesting, and a syntax I hadn't seen before.  We're looking into whether this breaks compatibility with cromwell, and thinking about whether this can/should be a broader change throughout our pipelines.  Just wanted to let you know that we're not ignoring this, just haven't gotten enough time to fully vet it yet!

serge2016:
@chrisamiller thank you! We are still working on it too:) There are several issues we have opened last time - they are consequences of our attempts to run the exome pipeline using `cwltool`...

tmooney:
Thanks again for the suggestion!  I've opened #774 that builds on this idea, so I'm going to close this PR.  Hopefully I'll get the Cromwell issues there sorted soon.

PR #687, opened 2019-05-29
Updated Mutect GATK3 with Mutect2 GATK4

saimukund20:


definitions/tools/mutect.cwl
```diff
@@ -2,75 +2,66 @@
...
+            set -o pipefail
+            set -o errexit
+
+            /gatk/gatk Mutect2 -O $1 -R $2 -I $3 -tumor $5 -I $4 -normal $6 -L $7
+            gunzip mutect.vcf.gz
+            sed "s/\t$5\t/\tTUMOR\t/g; s/\t$6/\tNORMAL/g" mutect.vcf > mutect_final.vcf
+            bgzip mutect_final.vcf
+            tabix mutect_final.vcf.gz
+            mv mutect.vcf.gz.stats mutect_final.vcf.gz.stats
+            /gatk/gatk FilterMutectCalls -R $2 -V mutect_final.vcf.gz -O mutect_final_filtered.vcf.gz
```
chrisamiller: It looks like FilterMutectCalls has lots of options that we may want to expose at some point, but the defaults seem reasonable for now.

definitions/pipelines/detect_variants.cwl
```diff
@@ -118,6 +118,10 @@ inputs:
     custom_clinvar_vcf:
         type: File?
         secondaryFiles: [.tbi]
+    tumor_sample_name_for_mutect2:
+        type: string
+    normal_sample_name_for_mutect2:
+        type: string
```
chrisamiller: Let's just call these "tumor_sample_name" and "normal_sample_name" and get them from the "tumor_name/normal_name" fields in upstream places like somatic_exome.cwl.  Alternately, I guess we could infer them from the header of the bam, since they're required to match the bam?

definitions/tools/mutect.cwl
```diff
@@ -2,75 +2,66 @@
...
+      - entryname: 'Mutect2.sh'
+        entry: |
+            set -o pipefail
+            set -o errexit
+
+            /gatk/gatk Mutect2 -O $1 -R $2 -I $3 -tumor $5 -I $4 -normal $6 -L $7
+            gunzip mutect.vcf.gz
+            sed "s/\t$5\t/\tTUMOR\t/g; s/\t$6/\tNORMAL/g" mutect.vcf > mutect_final.vcf
+            bgzip mutect_final.vcf
+            tabix mutect_final.vcf.gz
```
chrisamiller: You still need to add the steps we talked about, that will only apply the sample_name change to the appropriate header line.  Something like:

    grep "^##" mutect.vcf >mutect.final.vcf
    grep "^#CHROM" mutect.vcf | sed "s/\t$5\t/\tTUMOR\t/g; s/\t$6/\tNORMAL/g" >> mutect_final.vcf
    grep -v "^#" mutect.vcf >> mutect_final.vcf

chrisamiller: We should also think about whether it's reasonable to future-proof this to a time when we actually do use the sample names throughout, instead of "TUMOR" and "NORMAL".

definitions/pipelines/detect_variants.cwl
chrisamiller: The more I think about this, the more I think that you should grab those sample names from the headers of the bams. No matter what gets input into your sample name fields here, mutetech is going to base the headers of the VCF columns off of the sample names present in the header of the bam. If you want to be guaranteed that they will match, then you should pull them from the same place.

saimukund20:
This PR is to change Mutect (GATK3) to Mutect2 (GATK4). Here are the changes ive noticed:

1. Mutect2 doesn't require any of the inputs like dbsnp_vcf, cosmic_vcf , artifact_detection_mode , panel_of_normals_vcf , max_alt_alleles_in_normal_count , max_alt_allele_in_normal_fraction.
2. After just eyeballing the time taken to finish, Mutect2 is at least twice as fast as Mutect.
3. There were comparable differences with the sites called by Mutect2 compared to Mutect. Ive attached an image to show the similarities and differences of the Mutect VFCs and Final VCFs (Output of the somatic exome pipeline).
Mutect 3 vs 4

chrisamiller:
Adding that we reviewed the differences between the new and old final outputs, and they seem generally fine. They consist mostly of
- low-level VAF calls that could go either way.
- new dinucleotide variants (e.g. AG -> TG) that the new Mutect seems to be better at detecting

definitions/tools/mutect.cwl
```diff
@@ -2,75 +2,62 @@
...
+        entry: |
+            set -o pipefail
+            set -o errexit
+            
+            TUMOR=`(samtools view -H $3 | grep "^@RG" | sed "s/.*SM:\([^\t]*\).*/\1/g" | head -n 1)` #Extracting the sample name from the TUMOR bam.
+            NORMAL=`(samtools view -H $4 | grep "^@RG" | sed "s/.*SM:\([^\t]*\).*/\1/g" | head -n 1)` #Extracting the sample name from the NORMAL bam.
+            /gatk/gatk Mutect2 -O $1 -R $2 -I $3 -tumor $TUMOR -I $4 -normal $NORMAL -L $5 #Running Mutect2.
+            gunzip mutect.vcf.gz #Unzipping the Mutect2 output vcf file to alter the header names.
+            grep "^##" mutect.vcf >mutect.final.vcf #Extracting all the lines above #CHROM in the vcf to a new output vcf.
+            grep "^#CHROM" mutect.vcf | sed "s/\t$TUMOR\t/\tTUMOR\t/g; s/\t$NORMAL/\tNORMAL/g" >> mutect.final.vcf #Concatenating the #CHROM line with substituted sample headers to the output vcf.
```
tmooney: This is kind of an odd thing to have to do.  Is this because the new version of Mutect does something different from the old one?

definitions/tools/mutect.cwl
```diff
@@ -2,75 +2,62 @@
...
-        inputBinding:
-            prefix: "--cosmic"
-            position: 6
-        secondaryFiles: [.tbi]
-    artifact_detection_mode:
-        type: boolean?
-        inputBinding:
-            prefix: "--artifact_detection_mode"
-            position: 7 
-    panel_of_normals_vcf:
```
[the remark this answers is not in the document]
chrisamiller: It is probably not a bad idea - as outlined here, it will still accept a PON, although I think we're going to have to generate a new one:
 https://gatkforums.broadinstitute.org/gatk/discussion/10911/differences-between-gatk3-mutect2-and-gatk4-mutect2

> Mutect2 ignores sites present in the Panel of Normals (PoN) as well as sites that correspond to high fraction variants in the normal. By doing so, the tool avoids spending time in steps such as graph assembly and pairHMM alignments that cost compute. However, there is an option to force the tool to run the full process on sites that are in the PoN (--genotype-pon-sites), which can be useful in comparing results to older MuTect versions.
> 
> To create the PoN, call on each normal sample using Mutect2's tumor-only mode and then use GATK4's CreateSomaticPanelOfNormals, a tool new to GATK4. This contrasts with the GATK3 workflow, which uses an artifact calling mode in MuTect2 and CombineVariants for PoN creation. In GATK4, omitting to filter with FilterMutectCalls achieves the same result.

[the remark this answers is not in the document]
chrisamiller: Do we have a manifest of the samples used to create the PON?  Perhaps we should we round up some samples and generate a new one using some things sequenced on the NovaSeq?

definitions/tools/mutect.cwl
saimukund20: The new version of Mutect requires the VCF files to have the the sample names as the headers, cause it matches it from the BAM file.

definitions/tools/mutect.cwl
```diff
@@ -2,75 +2,62 @@
...
-        inputBinding:
-            prefix: "--max_alt_allele_in_normal_fraction"
-            position: 10
+
 outputs:
     vcf:
         type: File
         outputBinding:
-            glob: "mutect.vcf.gz"
+            glob: "mutect.final.filtered.vcf.gz"
```
apaul7: is the output for Mutect2 with GATK4 filtered?(I don't think this really matters as this specific file isn't one of the final output files)

saimukund20: Yes. So in GATK3 mutect, the filter step was inbuilt into that tool. Now its split into two steps. Mutect and then filtering.
https://gatkforums.broadinstitute.org/gatk/discussion/10911/differences-between-gatk3-mutect2-and-gatk4-mutect2

PR #688, opened 2019-05-30
updating to image with bugfixes

chrisamiller:
a few bugfixes related to t/n name swaps and handling of NA characters

### DIFF
--- a/definitions/tools/filter_vcf_somatic_llr.cwl
+++ b/definitions/tools/filter_vcf_somatic_llr.cwl
@@ -5,7 +5,7 @@ class: CommandLineTool
 label: "use the binomial/llr somatic filter to weed out low confidence variants"
 requirements:
     - class: DockerRequirement
-      dockerPull: mgibio/somatic-llr-filter:v0.4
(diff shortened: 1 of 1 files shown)


zlskidmore:
+1

definitions/tools/filter_vcf_somatic_llr.cwl
```diff
@@ -5,7 +5,7 @@ class: CommandLineTool
 label: "use the binomial/llr somatic filter to weed out low confidence variants"
 requirements:
     - class: DockerRequirement
-      dockerPull: mgibio/somatic-llr-filter:v0.3
+      dockerPull: mgibio/somatic-llr-filter:v0.4
```
tmooney: Do we want the very newest version?
```suggestion
      dockerPull: mgibio/somatic-llr-filter:v0.4.1
```

chrisamiller: weird - I thought I pushed that second commit through to update it...   Yes, thanks for catching that!

Issue #691, opened 2019-06-03
Replace symlinking with copying


tatahin:
Hello!
The following files contain `ln -s` command:
```
index_bam.cwl
index_cram.cwl
```
When running them with `cwltool` resulting output directory doesn't contain symlinks. I suggest to replace `ln -s` with `cp` for better compatibility with the reference implementation. What do you think about it?

susannasiebert:
The `ln -s` in the `pvac*.cwl`s doesn't pertain to symlinking output files. It's to symlink TMPDIR to a clean location because otherwise we were getting error of the TMPDIR path being too long in cromwell executions where the TMPDIR for each step is nested very deep in the output directory structure. Further down in the `arguments` you will see it calling the actual pvac* commands.

tatahin:
@susannasiebert, thank you! I hurried a bit after checking first two files... Updated the issue.

tmooney:
I've opened #692, which includes a proposed solution to this issue.

PR #692, opened 2019-06-03
Simplify BAM and CRAM indexers.

tmooney:
* We previously avoided the `InitialWorkDirRequirement` because a few executors didn't support it.  They generally all do now, so this is better than a manual copy.
* `cwltool` isn't propagating the symlinks, so make a copy of the index instead.  (Other executors end up making a copy later anyway.)  This addresses #691.

### DIFF
--- a/definitions/tools/index_bam.cwl
+++ b/definitions/tools/index_bam.cwl
@@ -4,9 +4,7 @@ cwlVersion: v1.0
 class: CommandLineTool
 label: "samtools index"
 arguments: [
-    "cp", $(inputs.bam.path), "$(runtime.outdir)/$(inputs.bam.basename)",
-    { valueFrom: " && ", shellQuote: false },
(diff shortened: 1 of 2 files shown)


chrisamiller:
+1

serge2016:
Could you explain, please, what is the purpose of:
```
      listing:
          - $(inputs.bam)
```
? Is it for checking the existence of input file?

tmooney:
@serge2016 In order to make the input BAM/CRAM available as an output along with the newly created index, the `InitialWorkDirRequirement` ensures that BAM/CRAM is placed in the working directory.   62b1331bb6f8268386b656506d2d7ac41dd11a61 switches to this standard CWL way of staging inputs instead of the manual `cp` we were employing before.  See https://www.commonwl.org/user_guide/15-staging/ for an example from the CWL User Guide.

serge2016:
@tmooney Thanks!!! Cool!

PR #698, opened 2019-06-04
Strip the secondary files from the input file to be indexed.

### LINKED ISSUE #697: Error: step index_bam cannot create regular file
Running* `exome_alignment.cwl` fails with massage:
```
[workflow alignment] starting step index_bam
[step index_bam] start
[job index_bam] Skipping Docker software container '--memory' limit despite presence of ResourceRequirement with ramMin and/or ramMax setting. Consider running with --strict-memory-limit for increased portability assurance.
[job index_bam] /tmp/ul288rou$ docker \
    run \
    -i \
    --volume=/tmp/ul288rou:/GouLsn:rw \
    --volume=/tmp/s18o4cg3:/tmp:rw \
    --volume=/tmp/cgjb835f/final.bam:/GouLsn/final.bam:ro \
    --volume=/tmp/cgjb835f/final.bai:/GouLsn/final.bai:ro \
    --workdir=/GouLsn \
    --read-only=true \
    --user=1000:1000 \
    --rm \
    --env=TMPDIR=/tmp \
    --env=HOME=/GouLsn \
    --cidfile=/tmp/ffpxvqkr/20190604143943-250533.cid \
    mgibio/samtools-cwl:1.0.0 \
    /bin/sh \
    -c \
    '/opt/samtools/bin/samtools' 'index' '/GouLsn/final.bam' '/GouLsn/final.bam.bai'  &&  'cp' 'final.bam.bai' '/GouLsn/final.bai'
cp: cannot create regular file '/GouLsn/final.bai': Read-only file system
[job index_bam] Max memory used: 0MiB
[job index_bam] completed permanentFail
[step index_bam] completed permanentFail
[workflow alignment] completed permanentFail
[step alignment] completed permanentFail
[workflow ] completed permanentFail
 ```
IMHO this is due to the .bai file, provided by tool `apply_bqsr.cwl`. So i tried to comment secondaryFiles in outputs.
```
outputs:
    bqsr_bam:
        type: File
        outputBinding:
            glob: $(inputs.output_name)
#        secondaryFiles: [^.bai]
```
After it exome_alignment.cwl works.

__________________________________________
*with `cwltool`

### PR DESCRIPTION
**Pull request by tmooney**
We don't ask for them, but `cwltool` sends them along to the output directory if a previous output included them. This change assumes the secondaryFiles are indexes and should be dropped so the main file can be re-indexed.

Closes #697.

### DIFF
--- a/definitions/tools/index_bam.cwl
+++ b/definitions/tools/index_bam.cwl
@@ -16,7 +16,8 @@ requirements:
       ramMin: 4000
     - class: InitialWorkDirRequirement
       listing:
-          - $(inputs.bam)
+        - ${ var f = inputs.bam; delete f.secondaryFiles; return f }
+    - class: InlineJavascriptRequirement
 inputs:
     bam:
         type: File
--- a/definitions/tools/index_cram.cwl
+++ b/definitions/tools/index_cram.cwl
@@ -16,7 +16,8 @@ requirements:
       ramMin: 4000
     - class: InitialWorkDirRequirement
       listing:
-          - $(inputs.cram)
+        - ${ var f = inputs.cram; delete f.secondaryFiles; return f }
+    - class: InlineJavascriptRequirement
 inputs:
     cram:
         type: File

### CODE REVIEW

**Comment by serge2016** [comment]
Are you sure, that it is a good solution?
Maybe it would be better to add a bash script (or `entry` CWL element) with checks:
```
...
baseCommand: ["/bin/bash", "check_and_gen_bais.sh"]
requirements:
      listing:
      - entryname: 'check_and_gen_bais.sh'
        entry: |
            set -o pipefail
            set -o errexit
            bam="$1"
            bai="$2"
            bambai="$1.bai"
            if [[ -s "$bai" && -s "$bambai" ]]; then
              echo "Both BAIs do exist."
            else
              if [ -s "$bai" ]; then
                cp "$bai" "$bambai"
              elif [ -s "$bambai" ]; then
                cp "$bambai" "$bai"
              else
                /opt/samtools/bin/samtools index $bam $bambai
                cp "$bambai" "$bai"
              fi
            fi
arguments:
    - position: 1
      valueFrom: "$(runtime.outdir)/$(inputs.bam.basename)"
    - position: 2
      valueFrom: "$(inputs.bam.nameroot).bai"
...
```

tmooney:
We were having a similar conversation earlier today.  The solution in this PR maintains how the pipelines behaved before the changes in #692.

In general I advocate for not calling `index_bam.cwl` after a tool that we know generates an index file.  Instead we should just copy the index to the other name as you suggest.

serge2016:
I think it is easier to modify the logic of `index_bam.cwl` tool and to rename it to `index_bam_if_needed.cwl` because I suppose, that there is no need to generate a new index if any already exists. That's why I think it would be better always to pass both `*.bam.bai` and `*.bai` files as optional secondary files to this tool and the tool should check:
if both `*.bam.bai` and `*.bai` exist, then skip (pass them through);
if at least one of them exists, then copy it to another;
in none of them exist, then create one and copy to another.

tmooney:
I'm going to merge this so that at least our code in master isn't broken. I've opened #738 to note the ongoing need to improve when the pipelines choose to run the indexer.

PR #701, opened 2019-06-12
Update depth-filter to 0.1.1

tmooney:
This has the fix for missing values that addresses genome/docker-depth-filter#1.


chrisamiller:
superceded by #708

PR #702, opened 2019-06-13
Add decompose step after the docm VCF gets merged in

susannasiebert:
Running GATK CombineVariants to merge in the DOCM VCF results in multi-allelic sites. Moving the `decompose` step to after the the `add_docm_variants` should fix this issue.


chrisamiller:
I believe the docm step requires a decomposed VCF, so I don't think this will work as intended. We'll need to decompose twice: once before docm, and once afterwards.  

It might make sense to leave the decompose step pre-docm, then add another decompose step as the last step of the docm subworkflow

chrisamiller:
I retract my statement - this should work fine!  +1

PR #708, opened 2019-06-17
bumping to 0.1.2 to remove PASS values when filtering

chrisamiller:
See here for the actual commit: 
https://github.com/genome/docker-depth-filter/commit/c14569ac8d88b4f226a77e975c913bb1f632f81c

using the built-in `add_filter` function correctly removes PASS values when filtering a line (instead of appending the filters and ending up with nonsense like "PASS,DEPTHFILTER"


chrisamiller:
Yeah, should have done a PR on that repo first 

This supercedes the fix made in https://github.com/genome/analysis-workflows/pull/701 so I'll close that one

Issue #659, opened 2019-05-14
Flip RNA-seq settings for Picard CollectRnaSeqMetrics.


jasonwalker80:
See internal JIRA BGAS-651 for more details, but the Picard strandedness settings are flipped.
https://github.com/genome/analysis-workflows/blob/ae75b938e6e8ae777a55686bbacad824b3c6788c/definitions/tools/generate_qc_metrics.cwl#L33

See this table for the correct settings:
https://github.com/griffithlab/rnaseq_tutorial/blob/master/manuscript/supplementary_tables/supplementary_table_5.md

Issue #658, opened 2019-05-14
Confirm the order of array inputs for the RNA-seq workflow


jasonwalker80:
With Toil it sounds like the order is reversed, but when running the same CWL with Cromwell we get a different order. This has impacted the RNA-seq QC metrics, see internal JIRA BGAS-651). Additionally, there will be a related issue forthcoming to correct the QC metric parameters for strandedness, they are reversed as well, see #659.

https://github.com/genome/analysis-workflows/blob/ae75b938e6e8ae777a55686bbacad824b3c6788c/definitions/subworkflows/bam_to_trimmed_fastq_and_hisat_alignments.cwl#L66

Most likely this fix in kallisto CWL will also need fixed:
https://github.com/genome/analysis-workflows/blob/23adee9d8b11679689660c6b50c3e5a40ea920f3/definitions/pipelines/rnaseq.cwl#L105

PR #711, opened 2019-06-18
Fixing strandedness settings in RNAseq pipeline

chrisamiller:
This fixes the errors outlined in issues https://github.com/genome/analysis-workflows/issues/658 and https://github.com/genome/analysis-workflows/issues/659, related to strandedness settings in QC, and to fastq order in Kallisto

### DIFF
--- a/definitions/pipelines/rnaseq.cwl
+++ b/definitions/pipelines/rnaseq.cwl
@@ -101,13 +101,7 @@ steps:
         in:
             kallisto_index: kallisto_index
             strand: strand
-            fastqs:
-                source: bam_to_trimmed_fastq_and_hisat_alignments/fastqs
-                valueFrom: |
-                    ${
-                      for(var i=0;i<self.length;i++){self[i] = self[i].reverse()}
-                      return(self)
-                     }
+            fastqs: bam_to_trimmed_fastq_and_hisat_alignments/fastqs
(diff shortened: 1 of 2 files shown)


definitions/pipelines/rnaseq.cwl
```diff
@@ -101,13 +101,7 @@ steps:
...
             kallisto_index: kallisto_index
             strand: strand
-            fastqs:
-                source: bam_to_trimmed_fastq_and_hisat_alignments/fastqs
-                valueFrom: |
-                    ${
-                      for(var i=0;i<self.length;i++){self[i] = self[i].reverse()}
-                      return(self)
-                     }
+            fastqs: bam_to_trimmed_fastq_and_hisat_alignments/fastqs
```
tmooney: Since the ordering can be problematic (apparently) depending on which engine is used, maybe we should switch `bam_to_trimmed_fastq_and_hisat_alignments.cwl` to use the specified read1 and read2 FASTQ outputs that the trimming step already provides?  Combining them into the right array of arrays here might still involve a little work, but it'd always be right!

chrisamiller: I agree, that'd be preferable, but I've got no time to figure that out at the moment - will file it away for a future date :)

PR #713, opened 2019-06-18
Basic Exome Alignment +QC Workflow (non-human)

jasonwalker80:
This replaces #610

### DIFF
--- a/definitions/pipelines/simple_exome_alignment.cwl
+++ b/definitions/pipelines/simple_exome_alignment.cwl
@@ -8,6 +8,7 @@ requirements:
       types:
           - $import: ../types/labelled_file.yml
     - class: SubworkflowFeatureRequirement
+    - class: ScatterFeatureRequirement
 inputs:
     reference: string
     bams:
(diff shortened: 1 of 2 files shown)


tmooney:
Does this need to be reconciled with <https://github.com/genome/analysis-workflows/blob/00313eea071a4199064a565dba2b2802fc9eb429/definitions/subworkflows/qc_exome_no_verify_bam.cwl> and the "mouse" pipeline <https://github.com/genome/analysis-workflows/blob/00313eea071a4199064a565dba2b2802fc9eb429/definitions/pipelines/exome_alignment_mouse.cwl>?

jasonwalker80:
I suppose `exome_alignment_mouse.cwl` is the same things... @chrisamiller agree? I had the PR open for a long time, but my forked repo was a mess. It looks like the same workflow except this produces BAM (which is fine given other workflows can still use the pipeline as a sub-workflow and compress to CRAM).

chrisamiller:
Yeah, based on a quick look, I believe the mouse exome alignment workflow does the same thing:
https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/exome_alignment_mouse.cwl

Essentially, it drops out bqsr and verify bam id to support non-human species, which is what you're doing here, seems like.

PR #716, opened 2019-06-24
Add aml_trio cwl pipeline

dufeiyu:
This is to replace the legacy aml_trio pipeline that is based on GMS Build-VariantReportingProcess.

Note that the docker image will be "mgibio/bam_readcount_helper-cwl:1.1.0" in bam_readcount.cwl once the PR https://github.com/genome/docker-bam_readcount_helper-cwl/pull/3 is merged.

### DIFF
--- a/definitions/pipelines/cle_aml_trio.cwl
+++ b/definitions/pipelines/cle_aml_trio.cwl
@@ -1,0 +1,614 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "Replace legacy AML Trio Assay"
+requirements:
+    - class: SchemaDefRequirement
+      types:
+          - $import: ../types/labelled_file.yml
+    - class: SubworkflowFeatureRequirement
+    - class: StepInputExpressionRequirement
+inputs:
+    reference: string
+    tumor_bams:
+        type: File[]
+    tumor_readgroups:
+        type: string[]
+    tumor_name:
+        type: string?
+        default: 'tumor'
+    normal_bams:
+        type: File[]
+    normal_readgroups:
+        type: string[]
+    normal_name:
+        type: string?
+        default: 'normal'
+    followup_bams:
+        type: File[]
+    followup_readgroups:
+        type: string[]
+    followup_name:
+        type: string?
+        default: 'followup'
+    mills:
+        type: File
+        secondaryFiles: [.tbi]
+    known_indels:
+        type: File
+        secondaryFiles: [.tbi]
+    dbsnp_vcf:
+        type: File
+        secondaryFiles: [.tbi]
+    bqsr_intervals:
+        type: string[]
+    bait_intervals:
+        type: File
+    target_intervals:
+        type: File
+    per_base_intervals:
+        type: ../types/labelled_file.yml#labelled_file[]
+    per_target_intervals:
+        type: ../types/labelled_file.yml#labelled_file[]
+    summary_intervals:
+        type: ../types/labelled_file.yml#labelled_file[]
+    omni_vcf:
+        type: File
+        secondaryFiles: [.tbi]
+    emit_reference_confidence:
+        type: string
+    gvcf_gq_bands:
+        type: string[]
+    intervals:
+        type:
+            type: array
+            items:
+                type: array
+                items: string
+    variant_reporting_intervals:
+        type: File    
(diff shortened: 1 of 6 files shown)


definitions/pipelines/gathered_cle_aml_trio.cwl
```diff
@@ -0,0 +1,288 @@
+#!/usr/bin/env cwl-runner
```
chrisamiller: just to make things line up nicely when looking through directories/lists of pipelines, it might be good to name this file "cle_aml_trio_gathered.cwl", so that it's right next to the original...

definitions/tools/bam_readcount.cwl
```diff
@@ -6,10 +6,12 @@ label: "run bam-readcount"
 
 baseCommand: ["/usr/bin/python", "/usr/bin/bam_readcount_helper.py"]
 requirements:
+    - class: ShellCommandRequirement
     - class: DockerRequirement
-      dockerPull: "mgibio/bam_readcount_helper-cwl:1.0.0"
+      dockerPull: "registry.gsc.wustl.edu/fdu/bam_readcount-prefix"
```
chrisamiller: Changing to an image hosted on the internal registry breaks all pipelines that use this tool for external users, so that seems not great.  Also, we should probably have a review of the changes to that bam_readcount_helper.py tool that you incorporated.

dufeiyu: In this PR comment, I mentioned: 
>Note that the docker image will be "mgibio/bam_readcount_helper-cwl:1.1.0" in >bam_readcount.cwl once the PR genome/docker-bam_readcount_helper-cwl#3 is merged.

definitions/pipelines/gathered_cle_aml_trio.cwl
dufeiyu: Hmm, somehow I have an impression that all gather cwls are named after gathered_xxxxxx.cwl

definitions/pipelines/cle_aml_trio.cwl
```diff
@@ -0,0 +1,614 @@
...
+            omni_vcf: omni_vcf
+            picard_metric_accumulation_level: picard_metric_accumulation_level   
+            minimum_mapping_quality: qc_minimum_mapping_quality
+            minimum_base_quality: qc_minimum_base_quality
+            final_name:
+                source: followup_name
+                valueFrom: "$(self).bam"
+        out:
+            [bam, mark_duplicates_metrics, insert_size_metrics, alignment_summary_metrics, hs_metrics, summary_hs_metrics, flagstats, verify_bam_id_metrics, verify_bam_id_depth]
+    concordance:
```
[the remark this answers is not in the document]
dufeiyu: Good point. So that means three way concordance run: tumor x normal, tumor x followup, normal x followup. Outputs of concordance.cwl will be overwritten unless a prefix kind of input need to be added to output filenames to differentiate in concordance.cwl.

definitions/pipelines/gathered_cle_aml_trio.cwl
chrisamiller: I think they are, but I'm saying that we should change that ;)

dufeiyu:
@chrisamiller I found that actually somalier can run on multiple bams and output proper paired relatedness accordingly. I made the change so aml_trio can get three samples' concordance.

definitions/tools/concordance.cwl
```diff
@@ -33,6 +33,11 @@ inputs:
         inputBinding:
             position: 4
         secondaryFiles: [.bai]
+    bam_3:
+        type: File?
+        inputBinding:
+            position: 5
+        secondaryFiles: [.bai]
```
chrisamiller: Would it be cleaner to pass this as an array of bams, since it can accept N samples?

dufeiyu: I was thinking of it too but not sure how secondary files .bai (required for somalier) can be passed along with corresponding bams if type: File[].

apaul7: to make sure they get passed through you should be able to do something like https://github.com/genome/analysis-workflows/blob/bcc6adaf15035f5ce6fc851e27b1173b0fd20c1c/definitions/tools/gatk_genotypegvcfs.cwl#L21-L28

```
type: File[]
```
is a shortcut for 
```
type:
  type: array
  items: File
```

PR #720, opened 2019-06-28
add annotsv.cwl tool

apaul7:
This PR adds the cwl tool for running AnnotSV. One unique thing with running AnnotSV is that it requires the path to the AnnotSV installation to be an enviornment variable. Found that the cwl requirement `EnvVarRequirement` allows me to set that.

Thought about having the AnnotSV installation aa part of the docker image but that would make the image kinda large and make it more challenging for adding custom annotations.

Future PR will add this and filtering to single sample sv caller sub workflow.

### DIFF
--- a/definitions/tools/annotsv.cwl
+++ b/definitions/tools/annotsv.cwl
@@ -1,0 +1,49 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+
+arguments: ["$(inputs.AnnotSV_path)/bin/AnnotSV", "-bedtools", "/usr/bin/bedtools", "-outputDir", "$(runtime.outdir)"]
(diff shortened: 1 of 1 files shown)


apaul7:
To add custom annotations you add a formatted bed file to a specific directory in the AnnotSV installation. Because of that I created a central spot for the AnnotSV files so that if we created any custom annotations everyone can use it. 

Thinking about this again a better approach for portability would be to have the AnnotSV in the docker image and then have optional input files for potential custom annotations. Creating a script which first puts those files in the correct spot and then runs AnnotSV.

tmooney:
That approach does sound better if we can swing it!  Then anyone using this pipeline won't need to also figure out how to set up AnnotSV itself.

apaul7:
made a new pr

Issue #715, opened 2019-06-19
Add SV call filtering to single sample sv caller subworkflow


apaul7:
Focus on filtering the Manta and Smoove callers using the paired and split read support

apaul7:
For Manta filtering:
Propose requiring 2+ split read support, 2+ paired read support and a 20% abundance of alternate reads for both paired and split reads. Have these values as the defaults but provide options to override them.
I have been able to get this working using bcftools but it requires updating to v1.9(https://github.com/samtools/bcftools/releases/tag/1.9)
The previous image has v1.3.1 but was unable to get the and/or logic to work correctly when variants do not contain the split read information but pass the paired read requirements.


Manta output fields:
```
##FORMAT=<ID=PR,Number=.,Type=Integer,Description="Spanning paired-read support for the ref and alt alleles in the order listed">
##FORMAT=<ID=SR,Number=.,Type=Integer,Description="Split reads for the ref and alt alleles in the order listed, for reads where P(allele|read)>0.999">
```
potential filtering expression:
```
bcftools filter manta.vcf \
-o manta_filtered.vcf \
-i '(SR[0:*]="." || (SR[0:1] >= 2 && (SR[0:1] / (SR[0:0]+SR[0:1]) > 0.2))) && (PR[0:1] >= 2 && (PR[0:1] / (PR[0:0]+PR[0:1]) > 0.2))'
```

`SR[0:*]="."` checks if `SR` field is not present
`SR[0:1] >= 2` checks if the number of alt split read support is greater or equal to 2
`SR[0:1] / (SR[0:0]+SR[0:1]) > 0.2` checks the abundance by dividing alt read count by the sum of ref and alt read count

`PR[0:1] >= 2` checks if the number of alt split read support is greater or equal to 2
`PR[0:1] / (PR[0:0]+PR[0:1]) > 0.2)` checks the abundance by dividing alt read count by the sum of ref and alt read count

The newer version of bcftools requires the `0:` syntax to use a sample. The `0` gets the first sample as they use 0 indexing. This shouldn't be an issue for the single sample sv caller subworkflow because there will be just 1 sample after the sv callers.


Examples:
variants that are filtered out
```
#CHROM 	POS 	ID 	REF 	ALT 	QUAL	FILTER	INFO	FORMAT	SAMPLE1
[...]

PR #721, opened 2019-06-28
add initial sv filtering tool

apaul7:
This PR adds the cwl tool for running bcftools to filter Manta and Snoove SVs based on issue #715 . 

Future PR will add this and AnnotSV to single sample sv caller subworkflow


definitions/tools/filter_sv_vcf.cwl
```diff
@@ -0,0 +1,87 @@
...
+        default: 0.2
+        inputBinding:
+            position: 4
+        doc: "required alternate paired read abundance percentage to pass"
+    split_count:
+        type: int?
+        default: 2
+        inputBinding:
+            position: 5
+        doc: " if present in variant, number of alternate split reads support needed to pass"
```
tmooney:
```suggestion
        doc: "if present in variant, number of alternate split reads support needed to pass"
```

definitions/tools/filter_sv_vcf.cwl
```diff
@@ -0,0 +1,87 @@
...
+          split_count="$5"
+          split_perc="$6"
+          vcf_source="$7"
+
+          if [ "$vcf_source" == "smoove" ]; then
+            echo "Running filter for smoove vcf"
+            filter_expression="(AS >= $split_count && (AS / (RS+AS) >= $split_perc)) && (AP >= $paired_count && (AP / (AP+RP) >= $paired_perc))"
+          elif [ "$vcf_source" ==  "manta" ]; then
+            echo "Running filter for manta vcf"
+            filter_expression="(SR[0:*]=\".\" || (SR[0:1] >= 2 && (SR[0:1] / (SR[0:0]+SR[0:1]) >= $split_perc))) && (PR[0:1] >= $paired_count && (PR[0:1] / (PR[0:0]+PR[0:1]) >= $paired_perc))"
```
tmooney: It looks like `$split_count` isn't used in the manta version.  I'd guess it goes here?
```suggestion
            filter_expression="(SR[0:*]=\".\" || (SR[0:1] >= $split_count && (SR[0:1] / (SR[0:0]+SR[0:1]) >= $split_perc))) && (PR[0:1] >= $paired_count && (PR[0:1] / (PR[0:0]+PR[0:1]) >= $paired_perc))"
```

definitions/tools/filter_sv_vcf.cwl
```diff
@@ -0,0 +1,87 @@
...
+requirements:
+    - class: ResourceRequirement
+      ramMin: 4000
+    - class: DockerRequirement
+      dockerPull: "mgibio/bcftools-cwl:1.9"
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: "filter_sv_vcf_read_support.sh"
+        entry: |
+          set -eou pipefail
```
tmooney: Not strictly necessary, but might be nice to have a `#!/bin/bash` line atop this file.

apaul7: done!

chrisamiller:
How were these default cutoffs decided upon?  I don't think they're necessarily bad, but it does seem like you're requiring both split and paired read support, which is fairly stringent, IIRC.

apaul7:
Dave suggested requiring 2 paired/split reads and the 20% abundance. The split reads are only required if there is a field for them in the call. 
With these filtering parameters I've been seeing ~90% of the calls being filtered out and ~10% passing.

PR #729, opened 2019-07-11
Added fusion calling into rnaseq workflow

johnegarza:
This PR adds support for fusion calling to the rnaseq workflow- pizzly, which makes calls based on fusion evidence from kallisto, and grolar, which parses pizzly output into a more readable format. Pizzly currently includes an option to specify the kmer size used to generate a kallisto index, defaulting to the kallisto default of 31 if not specified. Is this ever changed in our workflows? If not, it may be better to hard code this. Additionally, there are other pizzly options that we may want to expose, see [here.](https://github.com/pmelsted/pizzly) The grolar cwl contains an inline script for transparency and versioning; however, this may need some cleanup, and may require some updates from the [upstream script that it's based on](https://github.com/MattBashton/grolar) @chrisamiller


definitions/pipelines/immuno.cwl
```diff
@@ -50,6 +50,10 @@ inputs:
         type: File
     ribosomal_intervals:
         type: File
+    kallisto_kmer_size:
+        type: int?
```
chrisamiller: This param is configurable for kallisto, but not exposed in the CWL (intentionally, because we've never seen any reason to change it!) This implementation is a little problematic because it would allow the user to set this value for pizzly, but that wouldn't propagate through to kallisto, which could cause a mismatch.  

I think we suffer from parameter overload already, and this is one example of where we can make an opinionated choice for the end user.  My suggestion would be to hardcode the default value of 31 at the level of the pizzly and kallisto tools, but not expose it upstream. Happy to discuss if you disagree!

definitions/pipelines/immuno.cwl
```diff
@@ -305,6 +309,21 @@ outputs:
...
+        outputSource: rnaseq/unfiltered_fusions_json
+    filtered_fusion_seqs:
+        type: File
+        outputSource: rnaseq/filtered_fusion_seqs
+    filtered_fusions_json:
+        type: File
+        outputSource: rnaseq/filtered_fusions_json
+    parsed_fusion_calls:
+        type: File
+        outputSource: rnaseq/parsed_fusion_calls
```
chrisamiller: 1) naming making it clear which of these files is the final result would be nice.  (a naive user is not going to know whether filtered or parsed is the end stage).   

2) If this PR is needed soon, then this is not required, but this seems like as good a place as any to start putting things in subfolders?  A gene_fusions_pizzly folder containing these results would take a few files out of the already overwhelming list.

definitions/tools/grolar.cwl
```diff
@@ -0,0 +1,185 @@
...
+             # Test
+             #GetDistance(22, output)
+             #GetDistance(476, output)
+             #GetDistance(5916, output)
+           
+             # Get distances
+             cat("Computing gene distances\n")
+             geneDistance <- sapply(identical_idx, function(x) GetDistance(x, output))
+             output[identical_idx,"gene_distance"] <- geneDistance
+
```
chrisamiller: His revised code has this block wrapped in an if like:

    # Only run if our index has at least one TRUE value
    if(sum(identical_idx)) {
    
     # Get distances
      cat("Computing gene distances\n")
      geneDistance <- sapply(identical_idx, function(x) GetDistance(x, output))
      output[identical_idx,"gene_distance"] <- geneDistance
 
    }

That seems like a reasonable sanity check to add here

definitions/tools/pizzly.cwl
```diff
@@ -0,0 +1,56 @@
...
+class: CommandLineTool
+label: "pizzly gene fusion detector"
+baseCommand: ["pizzly"]
+requirements:
+    - class: InlineJavascriptRequirement
+    - class: ResourceRequirement
+      coresMin: 1
+      ramMin: 16000
+    - class: DockerRequirement
+      dockerPull: "chrisamiller/docker-pizzly"
```
chrisamiller: not urgent, but we should probably move this image out of my personal space and into mgibio.

definitions/pipelines/immuno.cwl
tmooney: Isn't letting the user supply their own kallisto index in effect exposing this parameter to them already?

chrisamiller: Shit, yeah I guess it is. so I guess we really do need to expose this and the right way to fix it is to add it as an input to the kallisto cwl (and up the chain) preferably with a default value of 31

definitions/tools/grolar.cwl
```diff
@@ -0,0 +1,185 @@
...
+           # Load JSON output from pizzly add in gene co-ordinates + compute distance when on same chr
+           
+           # Set up
+           # Change to your own output location
+           
+           # CRAN libs
+           library(jsonlite)
+           library(dplyr)
+           
+           # Assuming you have used GRCh38 gene models
```
apaul7: Adding the grolar step makes the whole rnaseq.cwl pipeline only compatible with GRCh38 references? If so shouldn't we make a note of this elsewhere? Or maybe some sort of check?

chrisamiller: Good point - that gets tricky, since you need to have the appropriate R package installed for each Ensembl version (and need to specify that as an input!).

One proposed solution:
1) this script should take additional arguments (ensembl version and species) from the pipeline
2) add some logic to convert that to the right package name 
3) It should check to see if that package is installed.  If not, it should install it in a user-configurable directory. 

So we would make `r_library_path` an input, and for our purposes, we'll set it to default to a path in `/core/` to use for these annotations.  The R script can then check whether this path exists at runtime.  If it does, it adds it to the library path. and proceeds to the check/install/run phase.  If the path does not exist, it can toss a warning and install the needed package to `/tmp/`. (which means that it will need to be installed anew every time that the tool is run).  Thoughts?

definitions/pipelines/immuno.cwl
johnegarza: Since the workflow/kallisto tool don't generate an index each time and instead accept a pre-existing one, does it make sense to add this upstream, since it won't do anything? I checked the kallisto source code and used it to make a small script that can pull the kmer size out of an index file. How do we feel about adding that to the workflow and dropping the kmer parameter entirely, since the index is always an input anyway?

chrisamiller: I like it!

definitions/tools/kmer_size_from_index.cwl
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Helper script to pull the k-mer size used in generating a kallisto index from the index file"
+baseCommand: ["g++", "-o", "extractor", "extractor.cpp"]
```
tmooney: This seems clever to be able to compile C++ on the fly, but should it be  compiled once into a docker image once instead of every time a workflow runs?

definitions/pipelines/immuno.cwl
tmooney: Now that this is determined inside `pipelines/rnaseq.cwl`, I believe this input can be removed here entirely.

definitions/tools/pizzly.cwl
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "pizzly gene fusion detector"
+baseCommand: ["pizzly"]
+requirements:
+    - class: InlineJavascriptRequirement
```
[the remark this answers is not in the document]
johnegarza: Good catch, no to both

definitions/tools/kmer_size_from_index.cwl
johnegarza: Yeah, this is a lot easier to read. Updating

Issue #717, opened 2019-06-25
Add support for FASTQ rather than unaligned BAM pipeline inputs


jasonwalker80:
Each DNA, RNA, and WGBS should also support paired, compressed FASTQ files as inputs. This changes the order and/or the commands used during adapter trimming and alignment. For example, we can no longer pipe interleaved FASTQ from the unaligned BAM directly to BWA-MEM. Subsequent Docker image changes/additions are required to update the actual commands executed in the docker shell script wrappers. Perhaps a new FASTQ compatible wrapper in addition to the existing unaligned BAM wrapper would help facilitate the transition.

chrisamiller:
This change will also need to include updating the example_data files and test yamls

jasonwalker80:
I'm going to close this since we added support for either FASTQ or unaligned BAM and most workflows utilize the new sequence type. If I missed something, let's create a more specific issue that can be addressed.

PR #730, opened 2019-07-11
A tool to align from FASTQs directly.

tmooney:
This is similar to the existing `align_and_tag.cwl` but without the `SamToFastq` pipe.  It's the first step to #717.

### DIFF
--- a/definitions/subworkflows/fastq_to_bqsr.cwl
+++ b/definitions/subworkflows/fastq_to_bqsr.cwl
@@ -2,7 +2,7 @@
 
 cwlVersion: v1.0
 class: Workflow
-label: "FASTQ to BQSR and VCF"
+label: "FASTQ to BQSR"
 requirements:
     - class: ScatterFeatureRequirement
     - class: SubworkflowFeatureRequirement
--- a/definitions/subworkflows/fastq_to_bqsr.cwl
+++ b/definitions/subworkflows/fastq_to_bqsr.cwl
@@ -1,0 +1,96 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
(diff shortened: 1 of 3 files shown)


definitions/tools/fastq_align_and_tag.cwl
```diff
@@ -0,0 +1,56 @@
...
+arguments:
+    - position: 5
+      valueFrom: $(runtime.cores)
+inputs:
+    fastq_1:
+        type: File
+        inputBinding:
+            position: 1
+        doc: 'fastq of sequence data (first read if paired end)'
+    fastq_2:
```
apaul7: Would having this as optional throw off the positions of the input variables for the bash script? and actually cause the bash script to fail because an input variable doesn't exist?

definitions/subworkflows/fastq_to_bqsr.cwl
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: Workflow
+label: "FASTQ to BQSR and VCF"
```
apaul7:
```suggestion
label: "FASTQ to BQSR"
```

definitions/tools/fastq_align_and_tag.cwl
tmooney: Yeah--you're right; the current script doesn't work right without a second FASTQ.  I'll just make it required for now; we can revisit if someone really wants to process fragment data.

definitions/subworkflows/fastq_to_bqsr.cwl
tmooney: Good point; updated `bam_to_bqsr.cwl` too :smile:

PR #735, opened 2019-07-15
adding annotsv.cwl

apaul7:
this PR adds the cwl tool for running AnnotSV. Decided to create a new branch/pr instead of pushing to #720 

Docker image now contains the AnnotSV files and the `ANNOTSV` env value necessary to run AnnotSV. This cwl does not expose the option to provide your own annotation files. Something that should be added in the future.

Future PR will add this and filtering to single sample sv caller sub workflow.

### DIFF
--- a/definitions/tools/annotsv.cwl
+++ b/definitions/tools/annotsv.cwl
@@ -30,11 +30,12 @@ inputs:
             prefix: "-outputFile"
         doc: "output file name"
     snps_vcf:
-        type: File?
+        type: File[]?
         inputBinding:
             position: 5
             prefix: "-vcfFiles"
-        doc: "snps vcf for adding hom/het snp counts found within svs"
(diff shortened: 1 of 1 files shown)


definitions/tools/annotsv.cwl
```diff
@@ -0,0 +1,43 @@
...
+        doc: "vcf file to filter"
+    output_tsv_name:
+        type: string?
+        default: "AnnotSV.tsv"
+        inputBinding:
+            position: 4
+            prefix: "-outputFile"
+        doc: "output file name"
+    snps_vcf:
+        type: File?
```
tmooney: Based on the docs, it looks like this could accept multiple files.

apaul7: yup you're right. the input binding is even -vcfFile**S**

PR #734, opened 2019-07-12
updating version, giving correct output name

chrisamiller:
this gets the job done for now. Issue with renaming of files noted here for future work: https://github.com/genome/analysis-workflows/issues/733

PR #736, opened 2019-07-16
Add a tool to logically rename a file and use it in our VCF filter subworkflow.

### LINKED ISSUE #733: renaming of files
It would be good to find a way to decouple the naming of files in tools from the names that are ouput by the workflows or subworkflows. 

As an example, `subworkflows/filter_vcf.cwl` currently returns the vcf file from whatever the last filter run is.  In order to make that sensible in the final ouput, I've renamed the output of `tools/filter_vcf_somatic_llr.cwl` to be `annotated_filtered.vcf`.  Really, though, if we want to keep our pipelines composable, it should return something like `somatic_filtered.vcf` and the the subworkflow should rename the final file appropriately.   

We talked through some suggestions in #cancer-genomics on slack, but I couldn't get it to work in the limited time I have today.  Specifically, Tom suggests: 

```
tmooney [12:50 PM]
Based on the previous example, I'd expect rename to work via `self[0].basename = 'annotated_filtered.vcf'` (edited) 
(If you did want a tool for renames, it could quite possibly be an `ExpressionTool` instead of a command-line tool.)

cmiller [12:54 PM]
that would go in the outputs section of the subworkflow?  something like this?
```outputs:  
   filtered_vcf:
        type: File
        outputSource: rename/renamed_file
        outputBinding: 
                OutputEval:  ${ self[0].basename = 'annotated_filtered.vcf' }```

tmooney [12:56 PM]
Yeah, something like that
(though I'm guessing the source won't be a step named `rename` :slightly_smiling_face:)
https://www.commonwl.org/v1.0/Workflow.html#CommandOutputBinding
by overriding the `basename` in the eval, it should rename the file theoretically--at least that was the suggestion in that github issue from earlier
I haven't tried it!

```

Also related to https://github.com/genome/analysis-workflows/issues/571 and https://github.com/genome/analysis-workflows/issues/712

tmooney:
This closes #733. It mostly supersedes #734.

The Expression tool could've been inline in the workflow, but this will hopefully be more convenient for re-using in other pipelines.

### DIFF
--- a/definitions/tools/rename.cwl
+++ b/definitions/tools/rename.cwl
@@ -2,6 +2,8 @@
 
 cwlVersion: v1.0
 class: ExpressionTool
+label: "Renamer"
+doc: "Logically renames a file, but not all workflow engines currently support it.  See also staged_rename.cwl"
 requirements:
     - class: InlineJavascriptRequirement
 inputs:
--- a/definitions/tools/rename.cwl
+++ b/definitions/tools/rename.cwl
@@ -1,0 +1,20 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: ExpressionTool
+requirements:
+    - class: InlineJavascriptRequirement
+inputs:
+    original:
+        type: File
+    name:
+        type: string
+outputs:
+    replacement:
+        type: File
+expression: |
+    ${
+        var f = inputs.original;
+        f.basename = inputs.name;
+        return {'replacement': f};
+    }
--- a/definitions/subworkflows/filter_vcf.cwl
+++ b/definitions/subworkflows/filter_vcf.cwl
@@ -5,6 +5,7 @@ class: Workflow
 label: "Apply filters to VCF file"
 requirements:
     - class: SubworkflowFeatureRequirement
+    - class: StepInputExpressionRequirement
 inputs:
     vcf:
         type: File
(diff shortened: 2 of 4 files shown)


tmooney:
@chrisamiller The name is currently set to:
https://github.com/genome/analysis-workflows/blob/0cddca7bf8a6b8c43ac3a0e6141b7ee6ed3b6c30/definitions/tools/filter_vcf_somatic_llr.cwl#L29
PR #734 is the one that includes the rename, so if that one is closed it will remain unchanged in master.

chrisamiller:
Ah, I see now - that never got merged.  That's what I get for reviewing code on my phone.  An enthusiastic +1!

tmooney:
Here's a JIRA issue I opened about Cromwell not supporting this: [BA-5835](https://broadworkbench.atlassian.net/browse/BA-5835).  [Selling your soul to Google required to access.]

tmooney:
I've added a `CommandLineTool`-based renamer.  If a workflow is run with `cwltool` I'd recommend staying with the original `rename.cwl`, but this one works more often :smile:

definitions/tools/rename.cwl
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env cwl-runner
+
```
chrisamiller: a comment or description mentioning that this doesn't work with cromwell would probably be a good idea, so we don't try to use it in workflows at present.

tmooney: Good point; I've added a `label` and `doc` to both tools.

PR #740, opened 2019-07-18
Introducing `sequence_data` type. Align either FASTQs or BAM!

tmooney:
Hot on the heels of #730 is a version that works on either FASTQ or BAM.  This approach saves us from having to maintain two versions of the pipeline and retains the efficiency of piping directly into `bwa` from BAM files.

If this approach looks good, I'll next replace the existing alignment steps in all the workflows and probably remove the existing alignment tool CWLs.

### DIFF
--- a/definitions/tools/sequence_align_and_tag.cwl
+++ b/definitions/tools/sequence_align_and_tag.cwl
@@ -12,9 +12,9 @@ requirements:
     - class: ResourceRequirement
       coresMin: 8
       ramMin: 20000
-      #    - class: DockerRequirement
-      #      dockerPull: "mgibio/alignment_helper-cwl:1.0.0"
-    - class: InlineJavascriptRequirement #necessary for cwltool to handle the record
+    - class: DockerRequirement
+      dockerPull: "mgibio/alignment_helper-cwl:1.0.0"
+    - class: InlineJavascriptRequirement
     - class: InitialWorkDirRequirement
       listing:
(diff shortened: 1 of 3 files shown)


definitions/tools/sequence_align_and_tag.cwl
```diff
@@ -0,0 +1,83 @@
...
+                        ;;
+                    n)
+                        NTHREADS="$OPTARG"
+                        ;;
+                esac
+            done
+
+            if [[ "$MODE" == 'fastq' ]]; then
+                /usr/local/bin/bwa mem -K 100000000 -t "$NTHREADS" -Y -p -R "$READGROUP" "$REFERENCE" "$FASTQ1" "$FASTQ2" | /usr/local/bin/samblaster -a --addMateTags | /opt/samtools/bin/samtools view -b -S /dev/stdin
+            fi
```
jasonwalker80: Should this be an else-if in case someone defined BAM and FASTQ inputs? Then the FASTQ would have precedence over the BAM (since it's more efficient).

tmooney: Only one of the two will ever happen.  In this bash script, whichever parameter is encountered last will determine the `$MODE`.

Further, the CWL type definition uses "exclusive" parameters, so it will error if both BAM and FASTQs are supplied.  (The two FASTQ parameters are also dependent so both must be supplied if one is.  See also [this user guide page](https://www.commonwl.org/user_guide/11-records/) for more details.)

PR #741, opened 2019-07-23
Use new sequence type in alignment.

tmooney:
This is #740 plus an implementation of that concept in the pipelines that include the standard alignment subworkflows.  I had to make an adapter subworkflow because Cromwell was not staging the file inputs when they were pulled as arguments directly leading to errors about files not being found.

I've tested this on the `somatic_exome.yaml` example but not run any full-scale pipelines yet.


tmooney:
My snarky answer would be "it shouldn't exist at all!", but I moved it under subworkflows 😄.

PR #743, opened 2019-07-23
adding cnvnator.cwl

apaul7:
This PR adds the CNVnator cwl as another copy number calling tool for WGS data. This is using the latest released version of `0.4`. This version of CNVnator does have CRAM and BAM support. Originally CNVnator required the reference to be split up in individual chromosome files. With the `0.4` version they now support compressed fasta.gz as a single file input. The issue with using that is that vcf generation perl script does not support a compressed reference input. I decided to generate the individual chromosome reference files when the tool is run. Allowing the CNVnator run and the cnvnator2VCF.pl script to use the same files. 

One issue that I have not been able to find a good solution for is that CNVnator requires the [ROOT](https://root.cern.ch) data analysis framework. In order to properly setup the environment a shell script needs to be sourced in the shell running CNVnator. Originally attempted to source this within the Dockerfile. Ran into a few issues where `source` is only found in a bash shell which is not the default for a Dockerfile. Then figured out that once the file is sourced it only modifies the shell that was used to run the sourcing command. I was able to add a line to the `~/.bashrc` file that would source the necessary ROOT shell script. This allowed CNVnator to work locally on my laptop but failed to work on the cluster. Then tried to add it to the script generated by the cwl. But the ROOT script has unset variables which causes the `run_cnvnator.sh` to fail because of the flags set. Ended up adding it as the first step in the `cnvnator.cwl`. Open to any suggestions or alternatives for this. 

One question I have is that I put the cnvnator2VCF.pl script on the docker image path. Should I be running `perl /path/to/cnvnator2VCF.pl $inputs` or is `cnvnator2VCF.pl $inputs` okay?

Have also been facing some issues with getting the docker image to build through dockerhub. I pushed my local image to `apaul7/cnvnator:0.4` for testing. Attempted to have dockerhub build the image but ran out of time. Tried pushing to `mgibio/cnvnator-cwl:0.4` to add the build cache hoping that would speed up future builds. But trying to rerun the build from the dockerfile hosted on github still failed. Currently splitting the Dockerfile in parts and running a build on the first sections to build up the build cache for subsequent builds.

Will keep this PR as a draft until the docker image is built successfully on `mgibio/cnvnator-cwl:0.4` or if it's okay to push a copy of the image I created locally. (I don't think it keeps the Dockerfile associated with the build like the automated builds do)

### DIFF
--- a/definitions/tools/cnvnator.cwl
+++ b/definitions/tools/cnvnator.cwl
@@ -2,7 +2,7 @@
 
 cwlVersion: v1.0
 class: CommandLineTool
-label: "Run CNVnator to calcualte copy number variations in WGS samples"
+label: "Run CNVnator to calculate copy number variations in WGS samples"
 
 arguments: ["source", "/opt/root/bin/thisroot.sh", { shellQuote: false, valueFrom: "&&" },  "/bin/bash", "run_cnvnator.sh"]
 
--- a/definitions/tools/cnvnator.cwl
+++ b/definitions/tools/cnvnator.cwl
@@ -58,7 +58,7 @@ inputs:
         default: 100
         inputBinding:
             position: 2
-        doc: "Bin size used to calculate covearge"
+        doc: "Bin size used to calculate coverage"
     chromosomes:
         type: string[]?
         default: ["chr1", "chr2", "chr3", "chr4", "chr5", "chr6", "chr7", "chr8", "chr9", "chr10", "chr11", "chr12", "chr13", "chr14", "chr15", "chr16", "chr17", "chr18", "chr19", "chr20", "chr21", "chr22", "chrX", "chrY"]
--- a/definitions/tools/cnvnator.cwl
+++ b/definitions/tools/cnvnator.cwl
@@ -1,0 +1,92 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Run CNVnator to calcualte copy number variations in WGS samples"
+
+arguments: ["source", "/opt/root/bin/thisroot.sh", { shellQuote: false, valueFrom: "&&" },  "/bin/bash", "run_cnvnator.sh"]
+
+requirements:
+    - class: DockerRequirement
+      dockerPull: "mgibio/cnvnator-cwl:0.4"
+    - class: ResourceRequirement
+      ramMin: 20000
+      coresMin: 1
+      tmpdirMin: 10000
+    - class: ShellCommandRequirement
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: "run_cnvnator.sh"
+        entry: |
+          #!/bin/bash
+          set -eou pipefail
+
+          # set vars
+          BAM="$1"
+          BIN_SIZE="$2"
+          CHROMOSOMES="${3//,/ }" # input as csv so it's one var to bash script but CNVnator wants them space separated
+          REFERENCE="$4"
+          SAMPLE="$5"
+
+          # create directory to store fasta files. CNVnator wants chromosomes in individual files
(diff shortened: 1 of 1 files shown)


apaul7:
The image has been created from the Dockerfile hosted on GitHub. I ended up downloading the ROOT binary instead of compiling it from source.
This pr is now ready for review

definitions/tools/cnvnator.cwl
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Run CNVnator to calcualte copy number variations in WGS samples"
```
tmooney:
```suggestion
label: "Run CNVnator to calculate copy number variations in WGS samples"
```

definitions/tools/cnvnator.cwl
```diff
@@ -0,0 +1,92 @@
...
+      listing:
+      - entryname: "run_cnvnator.sh"
+        entry: |
+          #!/bin/bash
+          set -eou pipefail
+
+          # set vars
+          BAM="$1"
+          BIN_SIZE="$2"
+          CHROMOSOMES="${3//,/ }" # input as csv so it's one var to bash script but CNVnator wants them space separated
```
tmooney: This works, but CWL should be able to handle passing a string containing spaces as a single parameter to the script if desired.

definitions/tools/cnvnator.cwl
```diff
@@ -0,0 +1,92 @@
...
+          REFERENCE="$4"
+          SAMPLE="$5"
+
+          # create directory to store fasta files. CNVnator wants chromosomes in individual files
+          # while with later versions of CNVnator these steps will accept a fasta.gz file the cnvnator2VCF.pl will not
+          mkdir FASTA_CHRS
+          awk 'BEGIN { CHROM="" } { if ($1~"^>") CHROM=substr($1,2); print $0 > "FASTA_CHRS/"CHROM".fa" }' "$REFERENCE"
```
tmooney: If we're going to run this regularly, should we create a directory of these in advance?  Some of our past reference sequence builds already contain per-chromosome FASTAs (e.g. GRCh37-lite-build37).

definitions/tools/cnvnator.cwl
```diff
@@ -0,0 +1,92 @@
...
+        type: File
+        inputBinding:
+            position: 1
+        doc: "BAM or CRAM input"
+    bin_size:
+        type: int?
+        default: 100
+        inputBinding:
+            position: 2
+        doc: "Bin size used to calculate covearge"
```
tmooney:
```suggestion
        doc: "Bin size used to calculate coverage"
```

definitions/tools/cnvnator.cwl
apaul7: I've thought about doing something like that. One reason I decided not to was that it would require a new input in the CWL that is either the directory path or a directory object. Or since our input reference is a string check that the directory where that file is located for split chromosomes directory? But that would break if reference is of type `File`.
Figured creating this on the fly would be the easiest way to implement it now. 

In the future changes could be made to the `cnvnator2VCF.pl` script that would allow it to handle a fasta.gz file. Then have the input to this step be a fasta.gz reference which would avoid having to create the individual chromosome files.

tmooney: Makes sense!  Yeah, that would end up requiring a whole new input were we not using a string.

definitions/tools/cnvnator.cwl
apaul7: yup you're right. Forgot that Cromwell puts quotes around inputs. Switched to 
```
itemSeparator: " "
```
and removed the modification of the input in the bash script. And it worked on my test data.

PR #744, opened 2019-07-24
Sanitize non-ACTGN bases in VCFs

chrisamiller:
This PR implements a "sanitize" step that is needed to remove non-IUPAC reference bases from VCF files. Currently, this occurs when a caller like varscan is run on the GRCh38 reference, which contains 95 positions with non [ACTGN] bases, all from the IUPAC standard (R, Y, S, B, W, etc).  The VCF spec, however, does not allow these bases, and GATK tools (correctly??) choke on them in a VCF.  

GATK-based tools (HaplotypeCaller, Mutect) just skip those non-ACTGN bases, and here, we replicate that behavior by removing any lines from VCFs that contain such a base.

I imagine that we haven't run into this before because this is the first time we're using the varscan-based tumor-only calling pipeline on whole genome data.   

Head of a validation experiment, showing some of the lines that are removed:
```
$ diff input.vcf sanitized.vcf  | cut -f 1-5
273792d273791
< chr1  248755122       .       R       A
357953d357951
< chr2  109493618       .       R       G
495844,495847d495841
< chr2  233142970       .       Y       T
< chr2  233143703       .       W       A
< chr2  233143709       .       Y       T
< chr2  233143729       .       Y       C
537613d537606
< chr3  66191221        .       Y       C
1465973d1465965
< chr7  154575459       .       R       A
. . . 
```

### DIFF
--- a/definitions/tools/vcf_sanitize.cwl
+++ b/definitions/tools/vcf_sanitize.cwl
@@ -1,0 +1,33 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Sanitize a VCF"
+baseCommand: ["/bin/bash","sanitize.sh"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 4
+      coresMin: 1
+    - class: DockerRequirement
+      dockerPull: "mgibio/samtools-cwl:1.0.0"
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: 'sanitize.sh'
+        entry: |
+            set -eou pipefail
+
+            # remove lines containing non ACTGN bases, as they conflict with the VCF spec
+            # and cause GATK to choke
+            gunzip -c "$1" | perl -a -F'\t' -ne 'print $_ if $_ =~ /^#/ || $F[3] !~ /[^ACTGNactgn]/' >sanitized.vcf
+            /opt/htslib/bin/bgzip sanitized.vcf
+            tabix sanitized.vcf.gz
+inputs:
+    vcf:
+        type: File
+        inputBinding:
+            position: 1
+outputs:
+    sanitized_vcf:
+        type: File
+        outputBinding:
+            glob: "sanitized.vcf.gz"
--- a/definitions/tools/vcf_sanitize.cwl
+++ b/definitions/tools/vcf_sanitize.cwl
@@ -6,7 +6,7 @@ label: "Sanitize a VCF"
 baseCommand: ["/bin/bash","sanitize.sh"]
 requirements:
     - class: ResourceRequirement
-      ramMin: 4
+      ramMin: 4000
       coresMin: 1
     - class: DockerRequirement
       dockerPull: "mgibio/samtools-cwl:1.0.0"
--- a/definitions/tools/vcf_sanitize.cwl
+++ b/definitions/tools/vcf_sanitize.cwl
@@ -15,12 +15,23 @@ requirements:
       - entryname: 'sanitize.sh'
         entry: |
             set -eou pipefail
-
-            # remove lines containing non ACTGN bases, as they conflict with the VCF spec
+            
+            # 1) removes lines containing non ACTGN bases, as they conflict with the VCF spec
             # and cause GATK to choke
-            gunzip -c "$1" | perl -a -F'\t' -ne 'print $_ if $_ =~ /^#/ || $F[3] !~ /[^ACTGNactgn]/' >sanitized.vcf
-            /opt/htslib/bin/bgzip sanitized.vcf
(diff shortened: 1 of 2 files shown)


definitions/tools/vcf_sanitize.cwl
```diff
@@ -0,0 +1,33 @@
...
+inputs:
+    vcf:
+        type: File
+        inputBinding:
+            position: 1
+outputs:
+    sanitized_vcf:
+        type: File
+        outputBinding:
+            glob: "sanitized.vcf.gz"
```
tmooney: What do you think of having the output file be named `$(inputs.vcf.nameroot).sanitized.vcf.gz`?

chrisamiller: Is there a straightforward way to strip the full `.vcf.gz` suffix?   `$(inputs.vcf.nameroot)`seems only to strip the final extension `.gz`.

definitions/tools/vcf_sanitize.cwl
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Sanitize a VCF"
```
johnegarza: Can we make the filename/label more specific? I'll link to the immuno PR as soon as I make it, but after our conversation a few days ago I also made a vcf sanitizing tool that fixes the illegal format tags (containing underscores) from mutect? Or maybe the two tools could be merged, since both handle issues that conflict with the VCF specification

definitions/tools/vcf_sanitize.cwl
tmooney: Ah yeah, the extension can have at most one `.` per the docs... Options are probably to split with JS or pass the whole basename into the shell script and let it staple the `sanitized` into the final filename.

definitions/tools/vcf_sanitize.cwl
chrisamiller: I agree with being more specific about what's going on here (I noted it below, but up top would be good too).  My vote is to merge the two scripts. Do you already have a PR for yours?

johnegarza: I'm working on the summary now, it'll be ready soon

johnegarza: See #751

chrisamiller:
@johnegarza  - This has been updated to add your mutect FORMAT field fixes, and to update the output basename.  Since it runs in the fp_filter (which operates on every snv/indel variant caller), it should take care of both the varscan and mutect issues.

definitions/tools/vcf_sanitize.cwl
```diff
@@ -0,0 +1,50 @@
...
+            echo "$1   $base    $outbase"
+            if [[ "$1" =~ ".gz" ]];then
+                #gzipped input
+                gunzip -c "$1" | perl -a -F'\t' -ne 'print $_ if $_ =~ /^#/ || $F[3] !~ /[^ACTGNactgn]/' | sed -e "s/ALT_F1R2/ALTF1R2/g;s/ALT_F2R1/ALTF2R1/g;s/REF_F1R2/REFF1R2/g;s/REF_F2R1/REFF2R1/g" >$outbase.sanitized.vcf
+            else 
+                #non-gzipped input
+                cat "$1" | perl -a -F'\t' -ne 'print $_ if $_ =~ /^#/ || $F[3] !~ /[^ACTGNactgn]/' | sed -e "s/ALT_F1R2/ALTF1R2/g;s/ALT_F2R1/ALTF2R1/g;s/REF_F1R2/REFF1R2/g;s/REF_F2R1/REFF2R1/g" >$outbase.sanitized.vcf
+            fi
+            /opt/htslib/bin/bgzip $outbase.sanitized.vcf
+            tabix $outbase.sanitized.vcf.gz
```
johnegarza: Does this work as is? I think it might need the `-p vcf` argument, the [documentation](http://www.htslib.org/doc/tabix.html) suggests that this expects gtf by default

definitions/tools/vcf_sanitize.cwl
```diff
@@ -0,0 +1,50 @@
...
+inputs:
+    vcf:
+        type: File
+        inputBinding:
+            position: 1
+outputs:
+    sanitized_vcf:
+        type: File
+        outputBinding:
```
johnegarza: Since we include a tabix step, I think we want a `secondaryFiles` field here too so that the index is passed along with the main file

chrisamiller: I still cannot really wrap my head around how secondaryFiles work, so I will take your word for it.  Added.

chrisamiller:
Comments addressed, runs fine on the test data

PR #750, opened 2019-07-31
adding cnv filtering

### LINKED ISSUE #749: add filtering for cnv callers
Should generate a cwl tool which allows filtering based on the read depth for copy number variation(cnv) callers. This would be different than the current proposed pr for sv filtering #721 and issue #715  which is used to results from tools that detect svs from discordant paired reads. 

in pr #743 the resulting CNVnator vcf file contains info field `natorRD`
```
##INFO=<ID=natorRD,Number=1,Type=Float,Description="Normalized RD">
```

the pipeline generated CNVkit vcf contains info field `FOLD_CHANGE`
```
##INFO=<ID=FOLD_CHANGE,Number=1,Type=Float,Description="Fold change">
```

Propose having default cutoff values for duplication at 1.25 and deletion at 0.75. Which would remove any sv that has a read depth between 0.75 and 1.25.

apaul7:
this pr adds the cnv filtering tool for CNVkit and CNVnator.
closes #749

### DIFF
--- a/definitions/tools/filter_sv_vcf_depth.cwl
+++ b/definitions/tools/filter_sv_vcf_depth.cwl
@@ -1,0 +1,73 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+
+baseCommand: ["/bin/bash", "filter_sv_vcf_depth.sh"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 4000
+    - class: DockerRequirement
+      dockerPull: "mgibio/bcftools-cwl:1.9"
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: "filter_sv_vcf_depth.sh"
+        entry: |
+          #!/bin/bash
+          set -eou pipefail
+          
+          INPUT_VCF="$1"
+          OUTPUT_VCF="$2"
+          DEL_DEPTH="$3"
+          DUP_DEPTH="$4"
+          VCF_SOURCE="$5"
+
+          if [ "$VCF_SOURCE" == "cnvnator" ]; then
+            echo "Running filter for cnvnator vcf"
+            filter_expression="(natorRD>$DUP_DEPTH | natorRD<$DEL_DEPTH)"
+          elif [ "$VCF_SOURCE" ==  "cnvkit" ]; then
+            echo "Running filter for cnvkit vcf"
+            filter_expression="(FOLD_CHANGE>$DUP_DEPTH | FOLD_CHANGE<$DEL_DEPTH)"
+          else
+            echo "vcf source: '$VCF_SOURCE' is not supported for SV filtering"
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


apaul7:
I've been using the `docker-interactive mgibio/bcftools-cwl:1.9` to test and filter my sv calls with no issue. It's really fast and most of the time they complete in a second or two.

cnvkit does provide the probes field which could be filtered on. and the CNQ and GQ which look to be based on the probe count. 
```
##INFO=<ID=PROBES,Number=1,Type=Integer,Description="Number of probes in CNV">
##FORMAT=<ID=GQ,Number=1,Type=Float,Description="Genotype quality">
##FORMAT=<ID=CNQ,Number=1,Type=Float,Description="Copy number genotype quality for imprecise events">
```
CNVnator does provide a few other options that could be filtered on but I wasn't sure if I should filter on them yet. 
```
##INFO=<ID=natorRD,Number=1,Type=Float,Description="Normalized RD">
##INFO=<ID=natorP1,Number=1,Type=Float,Description="e-val by t-test">
##INFO=<ID=natorP2,Number=1,Type=Float,Description="e-val by Gaussian tail">
##INFO=<ID=natorP3,Number=1,Type=Float,Description="e-val by t-test (middle)">
##INFO=<ID=natorP4,Number=1,Type=Float,Description="e-val by Gaussian tail (middle)">
##INFO=<ID=natorQ0,Number=1,Type=Float,Description="Fraction of reads with 0 mapping quality">
##INFO=<ID=natorPE,Number=1,Type=Integer,Description="Number of paired-ends support the event">
```
For my analysis I've been filtering out calls that have more than 10% reads with 0 mapping quality. But not sure if I'm losing true SVs by doing this. I am running a few samples that had cmas which detected large deletions/duplications through alignment and sv callers. Trying to see if 1) can detect the calls and 2) do the filtering steps remove the calls

chrisamiller:
Totally fine with merging this as step 1 - Further refinements can be added later!

definitions/tools/filter_sv_vcf_depth.cwl
```diff
@@ -0,0 +1,73 @@
...
+        default: "filtered_cnv.vcf"
+        inputBinding:
+            position: 2
+        doc: "output vcf file name"
+    deletion_depth:
+        type: double?
+        default: 0.75
+        inputBinding:
+            position: 3
+        doc: "Cutoff depth for deletions, less than passes filter"
```
tmooney:
```suggestion
        doc: "Depth for deletions must be less than this cutoff to pass the filter."
```

definitions/tools/filter_sv_vcf_depth.cwl
```diff
@@ -0,0 +1,73 @@
...
+    duplication_depth:
+        type: double?
+        default: 1.25
+        inputBinding:
+            position: 4
+        doc: "Cutoff depth for duplications, greater than passes filter"
```
tmooney:
```suggestion
        doc: "Depth for duplications must be greater than this cutoff to pass the filter."
```

apaul7:
Your suggestion does make it more clear!

PR #751, opened 2019-08-01
Immuno test

johnegarza:
This PR currently includes a rough draft of the changes that were needed to successfully complete a test run of the workflow. I wanted to get feedback before updating this PR and deciding what to do next. Summary:

1. Added a new tool, `sanitize_mutect_vcfs.cwl`. Mutect outputs several format tags that contain underscores, which are illegal according to the vcf spec and cause a needed GATK tool to fail. We might want to combine this with a similar tool created by @chrisamiller in #744 
2. Added a new tool, `replace_vcf_sample_name.cwl`, since renaming samples is required multiple times for different steps in the workflow. 
a. This is called back-to-back in lines 744 and 752 of `immuno.cwl`; it could be modified to accept multiple replacements so that it only requires one call here, or it could possibly be replaced by [bcftool reheader](https://samtools.github.io/bcftools/bcftools.html#reheader). If we stick to the custom tool, we can add a `tabix` step to the tool and remove all currently exposed index steps (lines 760, 775, 787). 
b. In addition, both of those steps don't actually allow inputs to the `sample_to_replace` field, they just use a default. This works with our current workflows, but the source of those default names should probably be pulled out from earlier in the workflow and passed in instead.
3. Added 3 new inputs, referring to the sample names in the bams/vcfs output by the `rnaseq`, `somatic`, and `germline` steps. 
a. It might be possible to pull `immuno_tumor_sample_name` and `immuno_normal_sample_name` directory from the preexisting `tumor_readgroups` and `normal_readgroups`. This could be done in `process_inputs.pl` for gms builds, or within the workflow for portability.
b. `immuno_tumor_sample_name_arr` is just `immuno_tumor_sample_name` wrapped in an array. Ideally, we should remove this input. This is only used as the newly added `samples_to_include` argument in `select_variants.cwl`; this was added as an array for flexibility and later use, but we could make the parameter a string instead to remove this issue. Alternatively, we could include code to coerce `immuno_tumor_sample_name` from a string to single element array somewhere in the workflow.

Other considerations:
- Should the steps from line 739 to the end of `immuno.cwl` be packaged together into a new subworkflow?
- There is significant redundancy happening by just calling both the somatic and germline exome workflow in `immuno.cwl`, since both workflows contain many of the same steps. Future versions of `immuno.cwl` may not call these pipeline workflows at all, just the portions needed; or they may call versions of them modified for use in this workflow. This is probably out of the scope of this PR.


definitions/pipelines/immuno.cwl
```diff
@@ -728,11 +735,65 @@ steps:
             optitype_name: optitype_name
         out:
             [cram,mark_duplicates_metrics,insert_size_metrics,insert_size_histogram,alignment_summary_metrics,hs_metrics,per_target_coverage_metrics,per_target_hs_metrics,per_base_coverage_metrics,per_base_hs_metrics,summary_hs_metrics,flagstats,verify_bam_id_metrics,verify_bam_id_depth,gvcf,final_vcf,coding_vcf,limited_vcf,vep_summary,optitype_tsv,optitype_plot]
+
+    sanitize_somatic_vcf:
```
chrisamiller: If we merge #744 and you rebase onto it, this step can now be removed.

definitions/tools/replace_vcf_sample_name.cwl
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Replace the sample name in a gzipped vcf and return"
+baseCommand: ['/bin/bash', 'replacer.sh']
```
chrisamiller: For portability, it might be nice to accept both gzipped and non-gzipped vcfs, since we have both at various points in our pipelines.

It also might be nice to just go ahead and index the vcf in here instead of having the additional index step in the calling workflow.

definitions/tools/replace_vcf_sample_name.cwl
```diff
@@ -0,0 +1,35 @@
...
+    - class: DockerRequirement
+      dockerPull: "mgibio/samtools-cwl:1.0.0"
+    - class: ResourceRequirement
+      ramMin: 8000
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: 'replacer.sh'
+        entry: |
+            set -eou pipefail
+            zcat $1 | sed -e "s/$2/$3/g" | /opt/htslib/bin/bgzip
```
chrisamiller:
```suggestion
            if [[ "$1" =~ ".vcf.gz" ]];then
                zcat "$1" | sed -e "s/$2/$3/g" | /opt/htslib/bin/bgzip >renamed.vcf.gz
            else
                cat "$1" | sed -e "s/$2/$3/g" | /opt/htslib/bin/bgzip >renamed.vcf.gz
            fi
            tabix renamed.vcf.gz
```

chrisamiller: would also need to update the glob (or pass in the output file name), but you get the gist

definitions/tools/sanitize_mutect_vcfs.cwl
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env cwl-runner
```
chrisamiller: as noted above, this can be removed when #744 is merged

definitions/pipelines/immuno.cwl
```diff
@@ -276,6 +276,13 @@ inputs:
     pvacseq_threads:
         type: int?
 
+    immuno_tumor_sample_name:
+        type: string
+    immuno_tumor_sample_name_arr:
+        type: string[]
+    immuno_normal_sample_name:
+        type: string
+
```
chrisamiller: I'm confused as to why these can't be the same as the sample names, and just passed in as "tumor_sample_name" and "normal_sample_name".  Those are easy to provide via `process_inputs.pl` if using GMS, or in the yaml if running outside it

susannasiebert: There is also `tumor_name` and `normal_name` although I'm not sure how they are used because they are set to `tumor` and `normal` by default. I agree that we should consolidate things wherever possible.

definitions/tools/replace_vcf_sample_name.cwl
susannasiebert: Right now this step is run twice in certain parts of the pipeline (once for the tumor and once for the normal sample name) so indexing after each step doesn't seem optimal.

susannasiebert:
We had also discussed that the phase_vcf.cwl subworkflow would need to be modified to add a step to only select the tumor sample from the somatic VCF. Did it turn out to not be necessary after all?

johnegarza:
@susannasiebert It's currently in the main immuno.cwl file, the step called "filter_somatic_vcf". I'm probably going to change the name and then move it from here into the subworkflow like we discussed.

susannasiebert:
@johnegarza ah, I see it now. Thanks for the pointer. I would suggest renaming that step to something like `select_tumor_sample` to be more descriptive.

johnegarza:
Closing and re-opening a new PR to account for revert & rework that was needed to fix issues

PR #753, opened 2019-08-08
Upgrade to version 1.5.0

susannasiebert:
This version adds AGFusion support to pVACfuse. AGFusion output is a directory, hence we now accept both `File` and `Directory` `type` for the `input_file`. 1.5.0 also adds a new tool, pVACbind, so this PR adds a CWL for it.

### DIFF
--- a/definitions/tools/pvacbind.cwl
+++ b/definitions/tools/pvacbind.cwl
@@ -1,0 +1,135 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "run pVACseq"
+
+baseCommand: [
+    "ln", "-s"
+]
+arguments: [
+    { valueFrom: "$TMPDIR", shellQuote: false },
+    "/tmp/pvacbind",
+    { valueFrom: " && ", shellQuote: false },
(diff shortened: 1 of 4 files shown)


definitions/tools/pvacbind.cwl
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "run pVACseq"
```
tmooney:
```suggestion
label: "run pVACbind"
```

definitions/tools/pvacbind.cwl
```diff
@@ -0,0 +1,135 @@
...
+    "--iedb-install-directory", "/opt/iedb",
+    { position: 5, valueFrom: $(runtime.outdir) },
+]
+requirements:
+    - class: ShellCommandRequirement
+    - class: DockerRequirement
+      dockerPull: "griffithlab/pvactools:1.5.0"
+    - class: ResourceRequirement
+      ramMin: 16000
+      coresMin: 8
```
tmooney: I'd either wire this up to the input:
```suggestion
      coresMin: $(inputs.n_threads)
```
or convert the `n_threads` input into an argument.

Issue #704, opened 2019-06-17
Output a summary TSV file from SV VCFs in WGS workflow


jasonwalker80:
Related to #545 and #544

apaul7:
the sv tsv is created by the AnnotSV cwl tool

PR #756, opened 2019-08-13
Update sv pipeline

apaul7:
This PR updates the single sample SV pipeline: 

- Adding CNVnator caller
- Adding filtering for SV/CNV callers
- Switch to AnnotSV for SV annotation and TSV generation
  Also required the snps detect variants to finish before starting the sv detect variants as the snps vcf is used in the annotation step. Maybe future optimizations can be done to have a detect variants cwl which runs the snp and sv detect variants(or better naming ideas?) at once
- remove `tools/filter_sv_vcf.cwl`
  which filtered by pop allele freq and geneated vcf files for input interval lists. AnnotSV generates a TSV with population allele frequency
- compress merged output vcf file
- compress and index filtered sv vcf files
- compress and index raw vcfs from cnvnator and cnvkit. smoove and manta already are compressed and indexed


apaul7:
This PR will also close issues #704 and #715

definitions/subworkflows/merge_svs.cwl
```diff
@@ -69,37 +50,19 @@ steps:
...
+            input_vcf: merge_sv_vcfs/merged_vcf
+            snps_vcf:
+                source: [snps_vcf]
+                valueFrom: ${ return [ self ]; }
         out:
-            [annotated_vcf, vep_summary]
-    sort_vcf:
-        run: ../tools/sort_vcf.cwl
+            [sv_variants_tsv]
+    merge_sv_bgzip:
```
tmooney: I might call this something like `bgzip_merged_sv_vcf`. The current name makes it sound like it is also performing a `merge` like the `merge_sv_vcfs` step above.

apaul7: I agree!

PR #757, opened 2019-08-14
Whole genome sequencing pipeline for Chipseq data including homer tag directory creation

saimukund20:


definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: 'Creating the homer tag directory'
+baseCommand: ["/bin/bash", "/gscmnt/gc2708/info/medseq/apl/chipseq/src/bam2homer.sh"]
```
chrisamiller: Probably a bad idea to link to files on the filesystem.  It'd be better to copy the script into the homer container, or just add it here as an inline bash script.

definitions/pipelines/Chipseq_alignment.cwl
```diff
@@ -0,0 +1,147 @@
...
+        type: File
+        secondaryFiles: [.tbi]
+    intervals:
+        type: File
+    per_base_intervals:
+        type: ../types/labelled_file.yml#labelled_file[]
+    per_target_intervals:
+        type: ../types/labelled_file.yml#labelled_file[]
+    summary_intervals:
+        type: ../types/labelled_file.yml#labelled_file[]
```
chrisamiller: What did we decide about these intervals? Is there a straightforward way to "blank" them so that we don't report target/base qc metrics? (since there's really no need for them in this pipeline?)

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,57 @@
...
+		set -o errexit
+
+		name=$1
+		bam=$2
+		reference_fasta=$3  #assumes that there's a .genome file next to the reference .fa
+		#run in chrisamiller/homer container
+		export PATH=$PATH:/opt/homer/bin
+		echo "name:     $name"
+		echo "bam:      $bam"
+		echo "reffasta: $reference_fasta"
```
chrisamiller: These three echo lines can be dropped - they were for troubleshooting

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,57 @@
...
+      listing:
+      - entryname: 'homer_tag_directory.sh'
+        entry: |
+		set -o pipefail
+		set -o errexit
+
+		name=$1
+		bam=$2
+		reference_fasta=$3  #assumes that there's a .genome file next to the reference .fa
+		#run in chrisamiller/homer container
```
chrisamiller: We should probably check for the presence of that genome file here and throw an informative error if it's missing.  (or possibly even require it as an input, like we do in some other pipelines)

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,57 @@
...
+		echo "name:     $name"
+		echo "bam:      $bam"
+		echo "reffasta: $reference_fasta"
+		if [[ ! -d $name/homer ]];then
+  			mkdir -p $name/homer
+		fi
+		echo "creating tagDir"
+		makeTagDirectory $name/homer $bam
+		echo "tagdir2Bed"
+		/gscuser/dspencer/programs/homer/bin/tagDir2bed.pl $name/homer/ | awk '$2>0 && $3>$2' | awk -v n=$name '{print >n"/"n".bed.part."$1}'
```
chrisamiller: This tagdir to bed stuff can all be dropped, AFAIK (lines 32-38). As long as we have the tagdir, we can run the downstream homer peak calling and such. I don't think I've ever actually used the tagdir bed file for anything

definitions/pipelines/Chipseq_alignment.cwl
```diff
@@ -0,0 +1,150 @@
...
+        outputSource: qc/per_target_hs_metrics
+    summary_hs_metrics:
+        type: File[]
+        outputSource: qc/summary_hs_metrics
+    bamcoverage_bigwig:
+        type: File
+        outputSource: qc/bamcoverage_bigwig
+    tag_directory:
+        type: Directory
+        outputSource: homer_tag_directory/tag_directory
```
chrisamiller: Is there any reason to make this dir name configurable? In the original bash script, it was because I was running multiple samples in the same directory, but in this case, every build is going to have it's own dir.  It doesn't hurt anything, I don't think, but might add some unnecessary complexity

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,55 @@
...
+      - entryname: 'homer_tag_directory.sh'
+        entry: |
+                set -o pipefail
+                set -o errexit
+
+                name=$1
+                bam=$2
+                reference_fasta=$3  #assumes that there's a .genome file next to the reference .fa
+                genome_file = `basename $3 .fa`
+                if [[ -f "$genome_file" ]];
```
chrisamiller: 1) `-s` might a better test to use here (check if the file exists and has non-zero size)
2) could be a little cleaner to use `if [[ ! -s "$genome_file" ]];then ...`  and then eliminate the first "does exist" block.  (just say if it doesn't exist, throw the error and die).

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: 'Creating the homer tag directory'
+baseCommand: ["/bin/bash", "homer_tag_directory.sh"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "chrisamiller/homer"
```
apaul7: can you use a tagged version of the docker image?

definitions/tools/homer_tag_directory.cwl
apaul7: Another thing to think about is that this will not work if we switch the reference input to type File

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,55 @@
...
+                    echo "File $genome_file does not exist. Please make sure that the .genome file is in the same directory as the .fa file."
+                    exit 1;
+                fi
+
+                export PATH=$PATH:/opt/homer/bin
+                if [[ ! -d $name/homer ]];then
+                    mkdir -p $name/homer
+                fi
+                echo "creating tagDir"
+                makeTagDirectory $name/homer $bam
```
apaul7: instead of modifying the path on line 31 why don't you write the full path `/opt/homer/bin/makeTagDirectory` here instead?

definitions/pipelines/Chipseq_alignment.cwl
```diff
@@ -0,0 +1,150 @@
...
+
+steps:
+    alignment:
+        run: ../subworkflows/bam_to_bqsr.cwl
```
apaul7: to help future proof this workflow could you use the `sequence_to_bqsr.cwl` subworkflow for alignment which contains the new sequence_data type?

definitions/tools/homer_tag_directory.cwl
chrisamiller: wouldn't be an awful idea to migrate this to an mgibio repo, so we can continue to use mine as a sandbox

saimukund20: Is there a tagged version of you image @chrisamiller ?

definitions/tools/homer_tag_directory.cwl
chrisamiller: adding to the PATH is needed here, because some of these homer tools call other tools internally and expect them to be in the PATH.

definitions/tools/bam_to_sam.cwl
```diff
@@ -0,0 +1,24 @@
...
+class: CommandLineTool
+label: 'BAM to SAM conversion'
+baseCommand: ["/opt/samtools/bin/samtools", "view", "-h", "-o"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "mgibio/samtools-cwl:1.0.0"
+    - class: ResourceRequirement
+      ramMin: 16000
+
+arguments: [out.sam]
```
tmooney: It might be nice to name this file based on the input file name like in `bam_to_cram.cwl`:
https://github.com/genome/analysis-workflows/blob/a23f42ef49c10a588fd35a3afaad5de03e253533/definitions/tools/bam_to_cram.cwl#L12

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,55 @@
...
+      listing:
+      - entryname: 'homer_tag_directory.sh'
+        entry: |
+                set -o pipefail
+                set -o errexit
+
+                name=$1
+                bam=$2
+                reference_fasta=$3  #assumes that there's a .genome file next to the reference .fa
+                genome_file = `basename $3 .fa`
```
tmooney: It says that it's looking for a `.genome` file, but where does the `.genome` part come from?  `all_sequences.fa` becomes `all_sequences` with the `basename` call, right?

chrisamiller: probably best just to require it as an input, (as we do in the bisulfite pipeline, IIRC)

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: 'Creating the homer tag directory'
+baseCommand: ["/bin/bash", "homer_tag_directory.sh"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "mgibio/homer:0.1"
```
chrisamiller: Not a requirement, but it might be nice to tag this with the homer release version instead (currently 4.10, I think)

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,43 @@
...
+                set -o errexit
+
+                name="homer_tag_directory"
+                bam=$1
+                reference_fasta=$2
+
+                export PATH=$PATH:/opt/homer/bin
+                if [[ ! -d $name/homer ]];then
+                    mkdir -p $name/homer
+                fi
```
chrisamiller: there's no point in seeing if the dir exists, since we know it doesn't exist up front in the CWL output dir - you can always just create it!

definitions/tools/homer_tag_directory.cwl
```diff
@@ -0,0 +1,43 @@
...
+      - entryname: 'homer_tag_directory.sh'
+        entry: |
+                set -o pipefail
+                set -o errexit
+
+                name="homer_tag_directory"
+                bam=$1
+                reference_fasta=$2
+
+                export PATH=$PATH:/opt/homer/bin
```
tmooney: If this is necessary for the proper operation of `homer`, it should be part of the Docker image.  In fact it already appears to be:
https://github.com/genome/docker-homer/blob/33adf12e1d5e1a49e8e12f72adb61352fa02d214/Dockerfile#L31

While it's harmless to add it a second time, it shouldn't be needed.

chrisamiller: The PATH sometimes doesn't get preserved in our interactive docker containers, which is why we have gotten into the habit of exporting the path in any script that uses homer.  If you can remove it and things still run properly, though, Sai, then sure, drop that line.

tmooney: On our cluster, you'd have to disable the envrionment preservation when running interactively.  This is included in the CWL pipeline runner:
https://github.com/genome/genome/blob/master/lib/perl/Genome/Model/CwlPipeline/Command/Run.pm#L235

definitions/tools/homer_tag_directory.cwl
saimukund20: Changed it

definitions/tools/homer_tag_directory.cwl
saimukund20: Changed it

definitions/tools/homer_tag_directory.cwl
saimukund20: I removed the export PATH line and it worked so I guess that solves the issue. I just had to add /opt/homer/bin in front of makeTagDirectory

definitions/pipelines/Chipseq_alignment.cwl
```diff
@@ -0,0 +1,148 @@
...
+        run: ../tools/bam_to_sam.cwl
+        in:
+            bam: alignment/final_bam
+        out: [final_sam]
+
+    homer_tag_directory:
+        run: ../tools/homer_tag_directory.cwl
+        in:
+            sam: bam_to_sam/final_sam
+            reference: reference
```
tmooney: Since the `reference` was removed from `homer_tag_directory.cwl`, it needs to be removed here, too.

saimukund20: Sorry about that! Just fixed it.

PR #759, opened 2019-08-14
make bqsr interval input optional

apaul7:
Cromwell fails to run if the bqsr interval is not provided in the input for alignment workflows.
Cromwell thinks the bqsr interval input is necessary because the sub workflow has it as required even though the above pipeline has it as optional and the tool called has a default value.


chrisamiller:
+1

Issue #761, opened 2019-08-15
add duphold depth annotations for manta/smoove calls 


apaul7:
* add a cwltool for running duphold, https://github.com/brentp/duphold
* update depth filtering tool to handled duphold annotated vcfs(use `DHFFC` for dels and `DHBFC` for dups), while also passing any non del/dup call
* add to single sample sv caller cwl subworkflow
   annotate manta and smoove calls
   filter dels/dups

apaul7:
```
DHFC: fold-change for the variant depth relative to the rest of the chromosome the variant was found on
DHBFC: fold-change for the variant depth relative to bins in the genome with similar GC-content.
DHFFC: fold-change for the variant depth relative to Flanking regions
```

apaul7:
the annotation with duphold and filtering have been added to the single sample sv caller sub workflow

PR #763, opened 2019-08-19
Add duphold tool

apaul7:
This cwl adds a tool for annotating SVs with [duphold](https://github.com/brentp/duphold). Providing coverage information for called deletions and duplications.
I added the threads option because the tool defaults to 4 threads while we default to 1 thread on the cluster.
First step in closing issue #761

### DIFF
--- a/definitions/tools/duphold.cwl
+++ b/definitions/tools/duphold.cwl
@@ -8,7 +8,7 @@ arguments: ["--threads", "$(runtime.cores)"]
 
 requirements:
     - class: ResourceRequirement
-      ramMin: 4000
+      ramMin: 10000
     - class: DockerRequirement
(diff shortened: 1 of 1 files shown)


definitions/tools/duphold.cwl
```diff
@@ -0,0 +1,53 @@
...
+        type: File
+        inputBinding:
+            position: 1
+            prefix: "--bam"
+        doc: "aligned bam/cram file"
+    output_vcf_name:
+        type: string?
+        default: "duphold_annotated.vcf"
+        inputBinding:
+            position: 4
```
tmooney: It's harmless, but just noting that two of the inputs have position `4`.

apaul7: Good catch. I moved the inputs around to get them in alphabetical order and didn't check/update the positions

PR #766, opened 2019-08-20
add bcftools merge cwl tool

apaul7:
This cwl tool allows an array of input vcfs to be merged together. input vcfs need to be bgzipped and tabix indexed. merge method `none` produces the best resulting vcf. using the method of `indels` or `both` causes the alternate field to get updated in a way where the variant has `alt1,alt2` which is not ideal for parsing downstream for SVs. The `indels` and `both methods also would only merge 2 calls if the variants have the same starting position and reference. With my testing only a very small number(~400 out of 32k) of variants from multiple callers fit this requirement and were being merged together.

### DIFF
--- a/definitions/tools/bcftools_merge.cwl
+++ b/definitions/tools/bcftools_merge.cwl
@@ -35,17 +35,26 @@ inputs:
             position: 3
             prefix: "--missing-to-ref"
         doc: "assume genotypes at missing sites are 0/0"
-    output_vcf_name:
-        type: string?
-        default: "bcftools_merged.vcf"
+    output_type:
+        type:
+            type: enum
+            symbols: ["b", "u", "z", "v"]
+        default: "z"
         inputBinding:
             position: 4
+            prefix: "--output-type"
+        doc: "output file format"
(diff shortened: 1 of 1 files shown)


definitions/tools/bcftools_merge.cwl
```diff
@@ -0,0 +1,54 @@
...
+inputs:
+    force_merge:
+        type: boolean?
+        default: true
+        inputBinding:
+            position: 1
+            prefix: "--force-samples"
+        doc: "resolve duplicate sample names"
+    merge_method:
+        type: string?
```
tmooney: This could be an `enum` of the possible valid merge methods.

apaul7:
I thought about exposing the `--output-type` option but that would require the `output_vcf_name` input to match the extension produced by bcftools. I guess since the `output_vcf_name` is an input the user can hopefully provide the correct file extension!

definitions/tools/bcftools_merge.cwl
apaul7: I agree that's the best for this input

PR #767, opened 2019-08-21
support non interleaved fastqs

apaul7:
The `-p` option is used for a single interleaved fastq file which gives this warning when aligning with 2 fastq files
```
[W::main_mem] when '-p' is in use, the second query file is ignored.
```
This doesn't cause the alignment step to fail. Downstream steps fail if they require paired end alignments.
The `-P` option should allow the 2 paired end fastq files to be aligned correctly. Manually rerunning this step with `-P` on one of my builds to confirm this works.

### DIFF
--- a/definitions/tools/fastq_align_and_tag.cwl
+++ b/definitions/tools/fastq_align_and_tag.cwl
@@ -25,7 +25,7 @@ requirements:
             REFERENCE="$4"
             NTHREADS="$5"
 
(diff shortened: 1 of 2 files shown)


apaul7:
Not sure. Originally I thought it was necessary to run the alignments but looks like it isn't. I don't have a lot of experience with running bwa. If you and others think `-P` is not needed I can remove it.

apaul7:
After reading into the options some more I do think it should be removed to make sure the pipeline produces the same alignments if you provide a bam or fastq files

chrisamiller:
+1

PR #771, opened 2019-08-22
add duphold depth filtering

apaul7:
Second step in closing #761 
Will filter vcfs based on a del/dup depth added by the duphold.cwl in #763 and pass the other sv calls. Currently Manta has other calls with svtype of `BND` and `INS` while Smoove produces calls with `BND` and `INV`


apaul7:
I'll add the parentheses to be consistent. For the expression I followed the example provided in the duphold README https://github.com/brentp/duphold#duphold-uphold-your-dup-and-del-calls

PR #772, opened 2019-08-22
Add sv size filter

apaul7:
This adds a cwl tool that allows you to filter svs based on the called size. Deletion calls have a negative number length while duplication calls have a positive number. Using the absolute value function allows you to filter both at the same time. 

I don't expect to use the `max` filter method but it was easy to add and there could be a future use I have yet to think of!

### DIFF
--- a/definitions/tools/filter_sv_vcf_size.cwl
+++ b/definitions/tools/filter_sv_vcf_size.cwl
@@ -21,12 +21,12 @@ requirements:
           SV_SIZE="$3"
           FILTER_METHOD="$4"
 
-          if [ "$FILTER_METHOD" == "max" ]; then
+          if [ "$FILTER_METHOD" == "max_len" ]; then
             echo "Running filter for max size svs"
-            filter_expression="ABS(SVLEN) < $SV_SIZE"
-          elif [ "$FILTER_METHOD" ==  "min" ]; then
+            filter_expression="ABS(SVLEN) <= $SV_SIZE"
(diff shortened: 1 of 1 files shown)


definitions/tools/filter_sv_vcf_size.cwl
```diff
@@ -0,0 +1,67 @@
...
+          OUTPUT_VCF="$2"
+          SV_SIZE="$3"
+          FILTER_METHOD="$4"
+
+          if [ "$FILTER_METHOD" == "max" ]; then
+            echo "Running filter for max size svs"
+            filter_expression="ABS(SVLEN) < $SV_SIZE"
+          elif [ "$FILTER_METHOD" ==  "min" ]; then
+            echo "Running filter for min size svs"
+            filter_expression="ABS(SVLEN) > $SV_SIZE"
```
tmooney: Do we want these to be inclusive ranges (i.e. `>=`)?  It might not make a huge difference either way; just not sure which way people would prefer to think about it.

apaul7:
What about switching to `max_size` and `min_size` and switching to inclusive ranges? Or use `max_len` and `min_len` to follow the `SVLEN` field.

tmooney:
Either of those names could be all right.  (I won't be terribly fussed if they stay as they are either; it might be I was overthinking it.)

apaul7:
I do think it should be clear what's going on. Don't want people to get confused or have to think a whole lot about what the tool does!

PR #774, opened 2019-08-22
Let reference input be either a string or a FIle

tmooney:
This is an updated version of the idea first presented in #682 (and potentially addresses #674).

I wanted to get this PR open, but we shouldn't merge this quite yet.  Using this version of the workflow requires a patch to Cromwell.  I have a change that makes it work as before when a `string` is supplied, but I'm still testing further changes to make Cromwell work correctly when a `File` is used instead.

### DIFF
--- a/definitions/tools/filter_vcf_mapq0.cwl
+++ b/definitions/tools/filter_vcf_mapq0.cwl
@@ -23,7 +23,10 @@ inputs:
             position: 2
         secondaryFiles: [.bai]
     reference: 
-        type: string
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai, ^.dict]
         inputBinding:
             position: 3
     threshold: 
--- a/definitions/tools/filter_vcf_mapq0.cwl
+++ b/definitions/tools/filter_vcf_mapq0.cwl
@@ -23,15 +23,15 @@ inputs:
             position: 2
         secondaryFiles: [.bai]
     reference: 
-         type: string
-         inputBinding:
-             position: 3
+        type: string
+        inputBinding:
+            position: 3
     threshold: 
-         type: float
-         inputBinding:
-             position: 4
+        type: float
+        inputBinding:
+            position: 4
 outputs:
-     mapq0_filtered_vcf:
-         type: File
-         outputBinding:
-             glob: "mapq_filtered.vcf.gz"
+    mapq0_filtered_vcf:
+        type: File
+        outputBinding:
+            glob: "mapq_filtered.vcf.gz"
--- a/definitions/tools/varscan_somatic.cwl
+++ b/definitions/tools/varscan_somatic.cwl
@@ -21,7 +21,10 @@ inputs:
             position: 2
         secondaryFiles: [^.bai]
     reference:
-        type: string
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai, ^.dict]
         inputBinding:
             position: 3
     strand_filter:
--- a/definitions/pipelines/alignment_exome.cwl
+++ b/definitions/pipelines/alignment_exome.cwl
@@ -10,7 +10,11 @@ requirements:
           - $import: ../types/sequence_data.yml
     - class: SubworkflowFeatureRequirement
 inputs:
-    reference: string
+    reference:
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai, ^.dict, .amb, .ann, .bwt, .pac, .sa]
     sequence:
         type: ../types/sequence_data.yml#sequence_data[]
     mills:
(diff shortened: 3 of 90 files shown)


definitions/tools/varscan_somatic.cwl
```diff
@@ -21,7 +21,10 @@ inputs:
             position: 2
         secondaryFiles: [^.bai]
     reference:
-        type: string
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai, ^.dict]
```
chrisamiller: Do all of these actually need the `.dict` file?  I don't relish the thought of digging through all 80-something of these to verify that, but I'm pretty sure, for example, that varscan here does not.

definitions/tools/filter_vcf_mapq0.cwl
```diff
@@ -23,15 +23,18 @@ inputs:
             position: 2
         secondaryFiles: [.bai]
     reference: 
-         type: string
-         inputBinding:
-             position: 3
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai, ^.dict]
```
chrisamiller: This is another one that definitely doesn't need the .dict file.

serge2016:
Thank you for implementing this!

tmooney:
@chrisamiller agreed, this was definitely a conservative approach to identifying which secondaryFiles to pass along.  Curiously, some workflow engines pass along all the secondary files from the original input, anyway, even if each tool doesn't explicitly request them, so it may be harder than expected to test which ones are needed.

I've just rebased this PR to account for it sitting a few months.  Now that I've got a new docker environment building it comes time where it'd be nice to sort this out.  Would you mind giving a re-review?  (Any other pipelines you'd like to see receive this treatment? For this PR I only focused on the main DNA pipelines that are all intertwined tool-wise.  It's generally safe for an unconverted pipeline to use a converted tool, though.)

tmooney:
I'd be happy to receive feedback on which `secondaryFiles` are needed. If there are any you're aware of off-hand, I could add it to this PR (along with the one you already mentioned).  However, I'm leaning towards getting this pushed and sorting it out more comprehensively in a follow-up.

(As long as we're using the `string` version of the input locally, what's listed is of zero consequence to our pipelines.  It's when we're running "in the cloud" that this would come into play.)

serge2016:
Dear colleagues, could you show an example of `*.index` file? You have added it as a secondary with reference fasta, but it doesn't exist in example data.

tmooney:
The `.index` file is generated by VEP:
https://useast.ensembl.org/info/docs/tools/vep/script/vep_cache.html#fasta

If you need to generate this `.index` file, this issue might be useful: Ensembl/ensembl-vep#211.

serge2016:
@tmooney, thank you! But those variants produce different `.index` files. Witch one is correct?
What should be in this `.index` file?

tmooney:
Sounds like it is better to use the htslib version rather than the older `Bio::DB::Fasta` version.  If nothing else I'd use an `.index` file produced by VEP itself after running it with a FASTA file; that's the most sure way to get one that will work.

It's functionally very similar to the `.fai` file--a quick way to find out where in the file each contig begins.  It needs to be in a format that VEP understands, which may vary depending on which modules are available to a particular install of VEP.  Thus it may be good to generate the file from within a Docker container running on the image used in the workflow, currently `mgibio/vep_helper-cwl:1.1.0`.

jasonwalker80:
I would add that our workflow is based on the implementation we have setup, ie. the `mgibio/vep_helper-cwl:1.1.0` docker image. The `.index` just happens to be the file that is generated when using that image. I agree the solution to the missing input is to run VEP once on the FASTA and let it generate the secondary file that is needed. As for the test data, we should include this step as an instruction to generated the secondary file before attempting to run the tests.

serge2016:
@tmooney, VEP doesn't produce `.index` file... That's a problem, I can't solve. VEP only produces `.fai`.

tmooney:
My apologies; I mixed up the two index formats.  The `.index` file comes from the older [`Bio::DB::Fasta`](https://metacpan.org/pod/release/CJFIELDS/BioPerl-1.5.9_1/Bio/DB/Fasta.pm) module rather than the newer version.  As you note, when htslib is available VEP prefers to use the newer index format instead and does not require or produce this `.index` file.

Perhaps the right course of action here is for us to remove the `.index` secondaryFile requirement from our pipelines.  As long as a new enough VEP with the proper modules is available, one could substitute an empty `.index` file for now.

serge2016:
@tmooney, thank you! For now, we use http://web.mit.edu/meme_v4.11.4/share/doc/fasta-make-index.html to generate `.index` file.

It would be great, if `.index` is removed from required files! Is it possible to make it optional secondary file?

tmooney:
@serge2016 PR #835 is open to remove this.  In CWL v1.1 there is a way to specify optional secondaryFiles (adding a `?` to the end of the extension), but we're not ready to migrate our workflows to this version yet.

PR #775, opened 2019-08-23
switch to total alt abundance filtering for sv read filtering

apaul7:
This pr modifies the paired/split read filtering for SVs by switching to a total alt abundance filter instead of paired alt abundance AND split alt abundance. 
Previously you could have a call with 10 split alt, 0 split ref, 2 paired alt, and 4 paired ref which would be filtered out because the paired alt abundance wouldn't meet the default 20% abundance requirement. 
With this new total alt abundance the call would pass
(10+2)/(10+0+2+4) = 75% 

Will need to wait till PR #756 is merged because this modifies the inputs.

### DIFF
--- a/definitions/pipelines/germline_wgs.cwl
+++ b/definitions/pipelines/germline_wgs.cwl
@@ -112,12 +112,10 @@ inputs:
         type: boolean
     merge_min_sv_size:
         type: int
-    sv_filter_paired_percentage:
+    sv_alt_abundance_percentage:
         type: double?
     sv_filter_paired_count:
         type: int?
-    sv_filter_split_percentage:
-        type: double?
     sv_filter_split_count:
         type: int?
     cnv_filter_deletion_depth:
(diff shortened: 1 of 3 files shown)


definitions/pipelines/germline_wgs.cwl
```diff
@@ -112,12 +112,10 @@ inputs:
         type: boolean
     merge_min_sv_size:
         type: int
-    sv_filter_paired_percentage:
+    sv_alt_abundance_percentage:
```
tmooney: Should this input still start with `sv_filter`? i.e.
```suggestion
    sv_filter_alt_abundance_percentage:
```

apaul7:
I thought about splitting some of the `or` steps to individual bcftools filter steps and then bcftools merge the resulting vcfs at the end. This would allow me to add a comment to explain what each step is filtering for. In the end I decided it would add more points of failure for the tool and not worth the trouble of having to debug failures.

PR #778, opened 2019-08-30
Switched vep arguments from string to enum for safety

johnegarza:


tmooney:
I'm not sure we'd want to limit VEP like this in the general case.  If the domain of values needs to be limited in the RNASeq pipeline, the value could possibly be converted out to a string for VEP using something like:
```cwl
        in:
            ensembl_assembly:
                source: vep_ensembl_assembly
                valueFrom: $(self)
```
(I've tried that with `cwltool` but not `cromwell`.)

chrisamiller:
This seems to represent a substantial hurdle to adding new genome builds (updating dozens of CWL files, instead of just pointing to a new reference and ensembl version).  I get the impetus, but I feel like there must be a better way...

PR #779, opened 2019-09-03
Bamrcnt insertion path

zlskidmore:
Workflow patch for the homozygous insertion bug in august
"insertions that were homozygous reference (usually in the normal) to receive 0/0 as their readcounts"

Patch allows for running the affected portion of the workflow in 5-10 minutes instead of rerunning the entire somatic_exome workflow from scratch.

Per @susannasiebert the code fix for the insertion bug was fixed in `bam_readcount_helper-cwl:1.1.1`


zlskidmore:
workflows has been tested to run successfully

definitions/tools/samtools_view_convert2bam.cwl
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env cwl-runner
```
chrisamiller: Any reason not to use the existing `cram_to_bam.cwl` instead of this? 
https://github.com/genome/analysis-workflows/blob/master/definitions/tools/cram_to_bam.cwl

tmooney:
I'd prefer to see example YAMLs all kept under `example_data` and not under `definitions`.  The `definitions` directory is what gets copied out when we make new processing profiles, so its better to keep it free of extra items.

If we do incorporate this into the repo, I might like to see it reframed as a `pipeline` instead of a patch.  Give it a nice name `readcount_and_filter_vcf.cwl` (or better) and it can be in our suite of things someone could use.

definitions/tools/samtools_view_convert2bam.cwl
zlskidmore: no reason other than I overlooked it when I went looking for it. I can create a one-off script to fill in the input yaml.... it would be a bit hacky but if everyones okay with that, would probably write a bash script to use the gms listers to retrieve the model data dirs and mv files from the patch over.

chrisamiller:
That might be nice - make it a vcf-filtering subworkflow, and swap out this bunch of steps in the `detect_variants.cwl` for one step - a call to this subworkflow

zlskidmore:
okay no problem if everyones in agreement I'll go ahead and formally break this into a sub-workflow.

I would argue to keep the cram_to_bam conversion in place and just accept that as another step.... I think it should run within a couple minutes if the input is already in bam format but we need it to patch things up

chrisamiller:
ugh, the need for bam-readcount to support crams strikes again... 

@tmooney Would it be possible to do something analogous to the "sequence" input that would allow these tools to take either cram or bam as input, converting if necessary?

tmooney:
@chrisamiller maybe.  The logic for any necessary conversions would have to be contained in a script--CWL has no concept of conditional execution (yet!).  Some options:
* A convert step that always runs but maybe does nothing.
* Including possible conversion (and a converter) in the script (and docker image) for any tool that cares.  (This is what the approach currently used for BWA to convert to FASTQ as needed.)

chrisamiller:
Handling conversion in each script, as is done in the sequence workflow, was my intent above, but I think I like the other idea even more. We could make a "prepare_sequence_inputs" step/script that takes a) the sequence data and b) the desired output format for that pipeline, then does whatever conversions are necessary and returns them.

zlskidmore:
I'm going to go ahead and close this, I have what should be a working script to take a model id, run the updated code, and put all the new results into a patched_results folder up one dir from results

I would rather just deprecate results in the results folder but I don't think that's possible given everything in there is owned by prod-builder and is read-only.

everything needed is contained here: /gscmnt/gc2565/zevallos_bga1684/zskidmor/homozygous_insertion_patch

Once my test finishes I'll make an announcement on slack

PR #781, opened 2019-09-18
Immuno workflow bugfixes

johnegarza:
This PR contains 2 small updates meant to fix bugs encountered while testing the workflow with new data. Grolar was updated to generate its output file in the correct location if there is no data for it to parse, and the mutect sample renaming logic was updated to account for tumor/normal pairs where the samples have similar names (in this case, HCC1395 and HCC1395_BL, respectively).

### DIFF
--- a/definitions/tools/mutect.cwl
+++ b/definitions/tools/mutect.cwl
@@ -24,7 +24,7 @@ requirements:
             /gatk/gatk Mutect2 -O $1 -R $2 -I $3 -tumor $TUMOR -I $4 -normal $NORMAL -L $5 #Running Mutect2.
             gunzip mutect.vcf.gz #Unzipping the Mutect2 output vcf file to alter the header names.
             grep "^##" mutect.vcf >mutect.final.vcf #Extracting all the lines above #CHROM in the vcf to a new output vcf.
-            grep "^#CHROM" mutect.vcf | sed "s/\t$TUMOR/\tTUMOR/g; s/\t$NORMAL/\tNORMAL/g" >> mutect.final.vcf #Concatenating the #CHROM line with substituted sample headers to the output vcf.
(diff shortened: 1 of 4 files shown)


definitions/tools/mutect.cwl
```diff
@@ -24,7 +24,7 @@ requirements:
             /gatk/gatk Mutect2 -O $1 -R $2 -I $3 -tumor $TUMOR -I $4 -normal $NORMAL -L $5 #Running Mutect2.
             gunzip mutect.vcf.gz #Unzipping the Mutect2 output vcf file to alter the header names.
             grep "^##" mutect.vcf >mutect.final.vcf #Extracting all the lines above #CHROM in the vcf to a new output vcf.
-            grep "^#CHROM" mutect.vcf | sed "s/\t$TUMOR/\tTUMOR/g; s/\t$NORMAL/\tNORMAL/g" >> mutect.final.vcf #Concatenating the #CHROM line with substituted sample headers to the output vcf.
+            grep "^#CHROM" mutect.vcf | sed "s/\<$TUMOR\>/TUMOR/g; s/\<$NORMAL\>/NORMAL/g" >> mutect.final.vcf #Concatenating the #CHROM line with substituted sample headers to the output vcf.
```
chrisamiller: Do the normal and tumor always have those angle bracket wraps? I don't recall those being there, but haven't looked recently

@saimukund20 can you look at this?

johnegarza: That's not actually part of the string, it's the syntax to tell sed to only replace exact matches. See https://stackoverflow.com/questions/26850806/sed-replace-exact-match

tmooney: I'd still prefer to see this replaced with something more robust as in my comment here: https://github.com/genome/analysis-workflows/pull/751#discussion_r315429275  It remains possible to have unusual sample names that thwart a regex-based approach with `sed`.

johnegarza: I agree; I went with the sed approach instead of a dedicated tool (like we did in another step in this workflow) because I think it would require breaking this script up, and I thought that might start to get out of the scope of this PR. Additionally, I'm planning to go through and refactor our workflows soon to remove all the steps that forcibly rename samples to TUMOR/NORMAL, so I can take care of it then.

PR #783, opened 2019-09-19
change filename to match output

zlskidmore:


zlskidmore:
I saw a cwl run fail at the grolar step, further investigation revealed it couldn't find the expected output. I took a quick look at the R code, I think this was just the result of a typo but correct me if i'm wrong

zlskidmore:
resolves issue #782

definitions/tools/grolar.cwl
```diff
@@ -79,7 +79,7 @@ requirements:
              output$geneA.id <- gsub(".\\d+$", "", output$geneA.id, perl = TRUE)
              output$geneB.id <- gsub(".\\d+$", "", output$geneB.id, perl = TRUE)
              cat(paste0("Writing out table: ", JSON_file, "_fusions_filt_sorted.txt", "\n"))
-             write.table(output[,c(-3,-4)], file = paste0(JSON_file, "_fusions_filt_sorted.simple.txt"), quote = FALSE, sep = "\t", row.names = FALSE, col.names = TRUE)
+             write.table(output[,c(-3,-4)], file = paste0(JSON_file, "_fusions_filt_sorted.txt"), quote = FALSE, sep = "\t", row.names = FALSE, col.names = TRUE)
```
chrisamiller: This function never gets called, as far as I can tell.  The lone call is to the `GetFusionz_and_namez` function, which does include the proper suffix.

johnegarza:
#781 includes a fix for this, not sure if you want to update your code and merge this in first or wait for mine to be included

definitions/tools/grolar.cwl
johnegarza: I've noticed that before as well, should we just remove it entirely? We could include it here or in my PR #781 which already has the fix for the file not found error

definitions/tools/grolar.cwl
```diff
@@ -100,7 +100,7 @@ requirements:
            
              if(length(output) < 1){
                print("No fusions exist in pizzly input. Creating empty output file")
-               file.create(paste0(JSON_file, "_fusions_filt_sorted.simple.txt"))
+               file.create(paste0(JSON_file, "_fusions_filt_sorted.txt"))
                quit(status=0)
```
chrisamiller: Yeah, this fix is the one you need, and makes sense

definitions/tools/grolar.cwl
zlskidmore: I missed the other function not being called, this line though in GenFusionz _and_namez is called and produces the error
https://github.com/genome/analysis-workflows/blob/ddd748516b25256a461ea9277303406fa2759b00/definitions/tools/grolar.cwl#L103

chrisamiller: I agree that it's confusing, but I think I'd like to minimize the differences from the maintainer's grolar script, so that we don't have to do a bunch of work if he releases new bugfixes or something. Maybe we could add comment or two?

zlskidmore:
okay so close this and wait for #781, is that the general consensus?

chrisamiller:
If you're in a hurry, drop the top change, keep the bottom and then we can approve it for merging.  John will have to rebase or merge when he gets the immuno PR fixed anyway

johnegarza:
I might still add more to my PR so it likely won't be in until early next week, so I agree that you should go ahead with yours if you're in a hurry

definitions/tools/grolar.cwl
```diff
@@ -100,7 +100,7 @@ requirements:
            
              if(length(output) < 1){
                print("No fusions exist in pizzly input. Creating empty output file")
-               file.create(paste0(JSON_file, "_fusions_filt_sorted.simple.txt"))
+               file.create(paste0(JSON_file, "_fusions_filt_sorted.txt"))
```
johnegarza: For this to end up in the correct Cromwell output directory, you need to add a `basename` wrapper like so: `file.create(paste0(basename(JSON_file), "_fusions_filt_sorted.txt"))`

zlskidmore:
okay that should do it, out of curiosity why does `GetFusionz` exist if it's not called?

chrisamiller:
Best I can tell, it's because grolar is a little bit hacky, but still useful (like most of the things I write, tbh!)

PR #784, opened 2019-09-24
Sv changes to wgs pipeline

apaul7:
This changes a few things and ended up being a bigger PR than I expected. 
- cnvkit subworkflow
  make default vcf name. previous tumor.nameroot expression does not work unless tumor bam is passed into the tool. Even though the tool itself doesn't use the input.
- remove the quotes around the CHROMOSOMES variable in the cnvnator shell script 
  was causing the input chromosomes to be parsed as a single chromosome 'chr1 chr2 chr3 ...' instead of 'chr1' 'chr2' 'chr3' ...
- update duphold docker image version 
  previous version had an issue where some manta duplications were not being annotated and then were being filtered out because they had no annotation.
- add bcftools sv merging and annotating with AnnotSV into the merge svs subworkflow
- move depth based sv caller filtering to a subworkflow
- move split/paired based sv caller filtering to a subworkflow
  still does the split/paired filtering
  added duphold annotating and filtering duplications and deletions based on expected coverage 
- add inputs/outputs to germline wgs pipeline
  add min sv size input to top workflow. Some users would like to remove the small variants called by the copy number callers and some do not.
- This PR also include changes from PR #775

### DIFF
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -24,6 +24,7 @@ inputs:
         type: File
     cnvkit_vcf_name:
         type: string?
+        default: "cnvkit.vcf"
     segment_filter:
         type:
           - "null"
@@ -73,16 +74,6 @@ steps:
             cns_file: cnvkit_main/tumor_segmented_ratios
             male_reference: male_reference
             cnr_file: cnvkit_main/tumor_bin_level_ratios
-            output_name: 
-                source: cnvkit_vcf_name
-                valueFrom: |
-                    ${  
-                        if(inputs.output_name) {
-                            return inputs.output_name;
-                        }   
-                        else {
-                            return inputs.tumor_bam.nameroot + ".cnvkit.vcf"
-                        }   
-                    }
+            output_name: cnvkit_vcf_name
         out:
             [cnvkit_vcf]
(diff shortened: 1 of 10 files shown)


definitions/subworkflows/cnvkit_single_sample.cwl
```diff
@@ -73,16 +74,6 @@ steps:
...
-                source: cnvkit_vcf_name
-                valueFrom: |
-                    ${  
-                        if(inputs.output_name) {
-                            return inputs.output_name;
-                        }   
-                        else {
-                            return inputs.tumor_bam.nameroot + ".cnvkit.vcf"
-                        }   
-                    }
```
jasonwalker80: Are other pipelines relying on a specified output file name? If not, this seems fine, but I'm curious why the prior behavior, ie. specfied output names.

definitions/tools/cnvnator.cwl
```diff
@@ -34,15 +34,15 @@ requirements:
           awk 'BEGIN { CHROM="" } { if ($1~"^>") CHROM=substr($1,2); print $0 > "FASTA_CHRS/"CHROM".fa" }' "$REFERENCE"
                     
           # extract read mapping from input bam(single sample)
-          cnvnator -root "${SAMPLE}.root" -tree "$BAM" -chrom "$CHROMOSOMES"
+          cnvnator -root "${SAMPLE}.root" -tree "$BAM" -chrom $CHROMOSOMES
```
tmooney: It might be better to have this script take the chromosomes at the end and
`shift 4` before using `"$@"` to pass the chromosome names along.  (This would entail no longer using the `itemSeparator` option in the CWL as well.)

definitions/subworkflows/cnvkit_single_sample.cwl
johnegarza: I think they were at the time this was written, but I don't see any place where it matters now, so it should be fine.

definitions/tools/cnvnator.cwl
apaul7: I created #792 and will work on updating then testing the cnvnator cwl tool as you suggested

PR #787, opened 2019-10-04
Adds single-end seq support to RNAseq pipeline

chrisamiller:


definitions/subworkflows/bam_to_trimmed_fastq_and_biscuit_alignments.cwl
```diff
@@ -22,35 +22,33 @@ inputs:
...
+        run: bam_to_trimmed_fastq.cwl
             bam: bam
-        out:
-            [fastq1, fastq2]
-    trim_fastq:
-        run: ../tools/trim_fastq.cwl
-        in:
-            reads1: bam_to_fastq/fastq1
-            reads2: bam_to_fastq/fastq2
+            paired_end: paired_end
```
tmooney: Looks like the `in:` key got lost somewhere along the way.

chrisamiller:
This should be ready to rock.  Review requested!

definitions/tools/hisat2_align.cwl
```diff
@@ -2,74 +2,122 @@
...
+                        strand="$OPTARG"
+                        ;;
+                    f)
+                        fastq_string="$OPTARG"
+                        ;;
+                esac
+            done
+
+
+            #collapse readgroups into separastring
```
tmooney: What's a `separastring`?

definitions/tools/kallisto.cwl
```diff
@@ -3,34 +3,46 @@
...
+            #fragment params are ignored and inferred from the data for paired-end
+            fragment_length="$5"
+            fragment_length_stddev="$6"
+
+            fastqs="${@:7}"
+
+            if [[ "$paired" == "true" ]];then
+              /usr/bin/kallisto quant -t $ncores -b 100 --fusion -o kallisto -i $index $strand $fastqs
+            else
+              ##"Reasonable" values for fragment length and stddev are hardcoded here. If your input data is substantially different
```
tmooney: This comment is confusing to me, since the below `kallisto quant` call has the value of `-b` hardcoded but not `-l` or `-s`.

definitions/tools/star_align_fusion.cwl
```diff
@@ -2,110 +2,125 @@
...
-arguments: [
-    "--runThreadN", $(runtime.cores)
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: 'star.sh'
+        entry: |
+            
+            set -eou pipefail
+
+            passthrough_args=${@:1:25}
```
tmooney: It's probably safer to have all the "passthrough args" at the end.  That way you don't have to be sure to count them properly.  For example, `out_samtype` looks like it may produce a variable number of parameters depending how many array values are used.

chrisamiller: Yeah, a better solution here is to use named parameters - I was trying to do it with minimal edits.  I'll work on that.

PR #789, opened 2019-10-09
add threads to merge command

### LINKED ISSUE #788: Fix merge_bams cwl to use multiple cores
samtools merge takes the `-@ nthreads`   argument to use multiple cores.  The number of cores are specified in the CWL, but not passed through to the program.
https://github.com/genome/analysis-workflows/blob/40097e1ed094c5b42b68f3db2ff2cbe78c182479/definitions/tools/merge_bams_samtools.cwl

apaul7:
closes #788


chrisamiller:
TIL that samtools now takes `--threads` in addition to the more cryptic `-@`.  

Thanks for doing this - I found it but hadn't had time to get back and fix it!

+1

PR #791, opened 2019-10-15
Remove bigwig generation from WGS QC process.

tmooney:
The tool is still left in place for now, in case anyone wants to use it for something else.

### DIFF
--- a/definitions/pipelines/alignment_wgs.cwl
+++ b/definitions/pipelines/alignment_wgs.cwl
@@ -93,9 +93,6 @@ outputs:
     summary_hs_metrics:
         type: File[]
         outputSource: qc/summary_hs_metrics
-    bamcoverage_bigwig:
-        type: File
-        outputSource: qc/bamcoverage_bigwig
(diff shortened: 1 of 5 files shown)


chrisamiller:
Why did this remove the tool from the chipseq cwl? (we were using that!) 

If it's a needed change, that's cool, but if not, we'd appreciate reverting it (and will switch to mosdepth or whatever when it becomes available)

tmooney:
It got removed only because it was taken out of `qc_wgs.cwl` and so was removed from anything using that.  The tool is still around, so it could be added separately to chipseq if that'd be helpful.

I suppose this is the downside of reusing components!

chrisamiller:
ah, i see now.  Yeah, we'll add it in as a separate step. Thanks for clarifying!

jasonwalker80:
Keep in mind the reason it was removed. It randomly failed on ~5% of samples. If that's ok, as Tom mentioned, feel free to add it back to chipseq.

PR #794, opened 2019-10-21
Parameterize custom vep annotations with custom type

apaul7:
This adds a new vep custom annotation type. Plan is to have this replace the individual inputs used with clinvar and gnomad vcfs and provide more customization to the pipelines. So when new versions of databases come out we can quickly add them to the pipelines.
I have gotten the `tumor_only_detect_variants.cwl` pipeline, which includes VEP annotation and gnomad filtering, to successfully complete. 

- To support the other file types I had to add the `.filter(String)` in the javascript parsing to remove the trailing comma.(non vcf input files do not have vcf fields as inputs!)
- for the somatic pipelines that do gnomad filtering, this assumes that you are adding the `AF` field and setting the gnomde_filtering field as true. and will then be filtering based on the `${annotation_name}_AF` (and if no gnomad input is provided defaults to VEP added gnomad frequencies)
- unsure how to provide the secondary files. the different data formats have different index files. I
 got around this by specifying the index file as a secondary file in the inputs 
an example input.yam for using the new VEP annotation type
```
vep_custom_annotations:
- method: 'exact'
  force_report_coordinates: true
  annotation:
    file:
      class: File
      path: gnomad-lite.genomes.r3.0.sites.vcf.gz
      secondaryFiles: [{class: File, path: gnomad-lite.genomes.r3.0.sites.vcf.gz.tbi}]
    data_format: 'vcf'
    name: 'gnomadw'
    vcf_fields: ['AF','AF_afr','AF_amr','AF_asj','AF_eas','AF_fin','AF_nfe','AF_oth','AF_sas']
    gnomad_filter: true
    check_existing: true
```

### DIFF
--- a/definitions/pipelines/detect_variants.cwl
+++ b/definitions/pipelines/detect_variants.cwl
@@ -105,6 +105,7 @@ inputs:
             - "null"
             - type: array
               items: ../types/vep_custom_annotation.yml#vep_custom_annotation
+              doc: "custom type, check types directory for input format"
 outputs:
     mutect_unfiltered_vcf:
         type: File
--- a/definitions/pipelines/detect_variants.cwl
+++ b/definitions/pipelines/detect_variants.cwl
@@ -101,11 +101,8 @@ inputs:
         type: string[]?
         default: [HGVSc,HGVSp]
     vep_custom_annotations:
-        type:
-            - "null"
-            - type: array
-              items: ../types/vep_custom_annotation.yml#vep_custom_annotation
-              label: "custom type, check types directory for input format"
+        type: ../types/vep_custom_annotation.yml#vep_custom_annotation[]
+        doc: "custom type, check types directory for input format"
 outputs:
     mutect_unfiltered_vcf:
         type: File
--- a/definitions/pipelines/detect_variants.cwl
+++ b/definitions/pipelines/detect_variants.cwl
@@ -338,6 +338,19 @@ steps:
             tumor_bam: tumor_bam
             do_cle_vcf_filter: cle_vcf_filter
             reference: reference
+            gnomad_field_name:
+              source: vep_custom_annotations
+              valueFrom: |
+                ${
+                   if  (typeof self !== 'undefined'){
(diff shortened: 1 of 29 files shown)


definitions/pipelines/detect_variants.cwl
```diff
@@ -97,12 +100,11 @@ inputs:
     vep_to_table_fields:
         type: string[]?
         default: [HGVSc,HGVSp]
-    custom_gnomad_vcf:
-        type: File?
-        secondaryFiles: [.tbi]
-    custom_clinvar_vcf:
-        type: File?
-        secondaryFiles: [.tbi]
+    vep_custom_annotations:
```
chrisamiller: adding a `doc` line for these inputs with some basic description of what this is expecting would be good (and/or a pointer to the custom types file, where you could add some comments about what/how/why)

apaul7: I've added labels to the different inputs in custom type yaml. Also added a doc line to all of the `vep_custom_annotations` input lines pointing people to the custom type yaml for the input format.

apaul7:
I've updated the example yamls. 

For the silent failing case I could leave the optional clinvar and gnomad vcf inputs in the workflows. This way the files are still annotated. Only issue would be figuring out how to do the gnomad filtering. 

If no `custom_vep_annotations` input is provided and a `custom_gnomad_vcf` is filter by the previous `gnomADe_AF` value? 
And if no `custom_vep_annotations` and `custom_gnomad_vcf` are provided filter by the vep provided gnomad frequencies(`gnomAD_AF`)?

tmooney:
If the new custom array input were made required, then it would force people to "upgrade" their old input YAMLs.  If no custom filtering is desired an empty array could be used.  (This required-but-empty-is-okay is how the various QC ROI file inputs currently work.)

apaul7:
Didn't know you could fulfill the required input by providing an empty array!
I've made the new custom annotation type a required input.

chrisamiller:
+1

PR #795, opened 2019-10-22
Immuno test data

johnegarza:
This PR contains some small changes that allow the test data set to run, as well as a new tool that accepts both optitype and (optional) clinical HLA typing calls, then produces a consensus call and mismatched call file.


definitions/tools/hla_consensus.cwl
```diff
@@ -0,0 +1,108 @@
...
+    - class: DockerRequirement
+      dockerPull: "python:3.7.4-slim-buster"
+    - class: ResourceRequirement
+      ramMin: 4000
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: 'hla_consensus.py'
+        entry: |
+                        import sys, os
+
```
chrisamiller: I looked through and get it, but could you add a comment here describing how the consensus is being generated, for the next time we have to look back at this code?

johnegarza:
After discussion, the consensus calling script is going to be more complicated than previously expected. I've removed it for now, to be included in its own PR following further discussion and modifications. This PR now only contains some small changes that allow the workflow test to complete successfully.

PR #801, opened 2019-10-29
adding default scatter count for mutect

chrisamiller:
Since scatter_count is a required input to the mutect subworkflow, it needs to either be a required input here or have a default value.  I chose the latter and set a default scatter of 50, which is what we provide upstream in the exome workflow anyway


chrisamiller:
tagging @saimukund20

PR #802, opened 2019-11-01
added star fusion detection

sridhar0605:
Added star fusion detection. 
1. changed rnaseq_star_fusion.cwl

TODO:
1. update yaml 
2. wiki create custom fusion genome index.


definitions/process_inputs.pl
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env genome-perl
```
chrisamiller: As of right now, these files don't go into the github repo, since they're specific to our GMS setup

PR #805, opened 2019-11-04
star_fusion_detect

sridhar0605:
Added star fusion detection.
1. changed rnaseq_star_fusion.cwl
TODO:
1. move genome_index to core/
2. update example yaml
3. wiki create custom fusion genome index.

### DIFF
--- a/definitions/pipelines/rnaseq_star_fusion.cwl
+++ b/definitions/pipelines/rnaseq_star_fusion.cwl
@@ -80,6 +80,7 @@ outputs:
         outputSource: generate_qc_metrics/metrics
     chart:
         type: File
+        outputSource: generate_qc_metrics/chart
     fusion_evidence:
         type: File
         outputSource: kallisto/fusion_evidence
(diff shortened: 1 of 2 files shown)


definitions/pipelines/rnaseq_star_fusion.cwl
```diff
@@ -14,6 +14,8 @@ inputs:
         type: string[]
     star_genome_dir:
         type: Directory
+    star_fusion_genome_dir:
+        type: Directory
```
chrisamiller: Can you add a quick "doc" line here (and to the star_genome_dir) line, giving a little info about what is expected to be in there?  (Or probably just linking to the wiki page for how to generate it!)

definitions/tools/star_fusion_detect.cwl
```diff
@@ -0,0 +1,60 @@
...
+  star_fusion_genome_dir:
+    type: Directory
+    inputBinding:
+        position: 2
+        prefix: '--genome_lib_dir'
+    doc: '
+    fusion genome dir
+    '
+
+  junction_file:
```
johnegarza: Can you add a short doc line here explaining what this is?

sridhar0605: Added doc

PR #807, opened 2019-11-06
Remove sample renaming

johnegarza:
This PR modifies the detect_variants* pipelines to preserve the actual sample names in all vcf headers, instead of always renaming them to NORMAL/TUMOR. Most of the changes to each file are the addition of 2 new arguments, `tumor_sample_name` and `normal_sample_name`, and the needed connections. Along the way, I moved the scripts of several tools from docker images into the CWL file itself. For those tools that needed changes to support the new naming scheme, transcription into CWL and the actual change to preserve the names are two separate commits.

### DIFF
--- a/definitions/pipelines/panel_of_normals.cwl
+++ b/definitions/pipelines/panel_of_normals.cwl
@@ -25,8 +25,6 @@ inputs:
         secondaryFiles: [.tbi]
     normal_sample_name:
         type: string
-    tumor_sample_name:
-        type: string
 outputs:
     pon_vcf:
         type: File
@@ -49,7 +47,7 @@ steps:
                 default: true
             panel_of_normals_vcf:
                 default: null
-            tumor_sample_name: tumor_sample_name
+            tumor_sample_name: normal_sample_name
         out:
             [unfiltered_vcf]
     combine:
--- a/definitions/pipelines/panel_of_normals.cwl
+++ b/definitions/pipelines/panel_of_normals.cwl
@@ -23,6 +23,10 @@ inputs:
     cosmic_vcf:
         type: File?
         secondaryFiles: [.tbi]
+    normal_sample_name:
+        type: string
+    tumor_sample_name:
+        type: string
 outputs:
     pon_vcf:
         type: File
@@ -45,6 +49,7 @@ steps:
                 default: true
             panel_of_normals_vcf:
                 default: null
+            tumor_sample_name: tumor_sample_name
         out:
             [unfiltered_vcf]
     combine:
--- a/definitions/tools/add_strelka_gt.cwl
+++ b/definitions/tools/add_strelka_gt.cwl
@@ -5,11 +5,99 @@ class: CommandLineTool
 label: "add GT tags"
 requirements:
     - class: DockerRequirement
-      dockerPull: "mgibio/cle:v1.3.1"
+      dockerPull: "ubuntu:xenial"
     - class: ResourceRequirement
       ramMin: 4000
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: 'add_strelka_gt.pl'
+        entry: |
+            #!/usr/bin/perl
+
+            use strict;
+            use warnings;
+
+            use feature qw(say);
+
+            die("wrong number of arguments") unless scalar(@ARGV) == 2;
+            my ($strelka_vcf, $outdir) = @ARGV;
+
+            open(my $strelka_vcf_fh, '-|', '/bin/gunzip', '-c', $strelka_vcf) 
+                or die("couldn't open $strelka_vcf to read");
+            open(my $add_gt_fh, ">", "$outdir/add_gt.vcf") 
+                or die("couldn't open add_gt.vcf for write");
+
+            while (<$strelka_vcf_fh>) {
+                chomp;
+                if (/^##/) {
+                    say $add_gt_fh $_;
+                }
+                elsif (/^#/) { #COLUMN HEADER
+                    say $add_gt_fh '##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">';
+                    say $add_gt_fh $_;
+                }
+                else {
+                    my @columns = split /\t/, $_;
+                    my ($ref, $alt, $info) = map{$columns[$_]}(3, 4, 7);
+                    my @alts = split /,/, $alt;
+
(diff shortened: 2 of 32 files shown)


definitions/pipelines/panel_of_normals.cwl
```diff
@@ -45,6 +49,7 @@ steps:
                 default: true
             panel_of_normals_vcf:
                 default: null
+            tumor_sample_name: tumor_sample_name
```
tmooney: The `tumor_bam` input has `normal_bams` as its value.  Should this also be `normal_sample_name` to match?  (As an additional note, this tool now has both `tumor_sample_name` and `normal_sample_name` inputs but only one gets used.)

chrisamiller:
Not done reviewing it yet, but one thing that we need to check is whether all these tools handle sample names with special characters.  I don't think testing an exhaustive list of ascii chars is necessary, but I know for a fact that spaces, single quotes and backslashes all exist in sample names in our database

definitions/tools/add_strelka_gt.cwl
```diff
@@ -5,11 +5,99 @@ class: CommandLineTool
 label: "add GT tags"
 requirements:
     - class: DockerRequirement
-      dockerPull: "mgibio/cle:v1.3.1"
+      dockerPull: "ubuntu:xenial"
```
tmooney: Is it time to use `bionic` instead? :smile:

definitions/tools/mutect.cwl
```diff
@@ -22,14 +22,7 @@ requirements:
             TUMOR=`(samtools view -H $3 | grep "^@RG" | sed "s/.*SM:\([^\t]*\).*/\1/g" | head -n 1)` #Extracting the sample name from the TUMOR bam.
             NORMAL=`(samtools view -H $4 | grep "^@RG" | sed "s/.*SM:\([^\t]*\).*/\1/g" | head -n 1)` #Extracting the sample name from the NORMAL bam.
             /gatk/gatk Mutect2 --java-options "-Xmx20g" -O $1 -R $2 -I $3 -tumor $TUMOR -I $4 -normal $NORMAL -L $5 #Running Mutect2.
```
chrisamiller: This chunk worries me a little bit.  There could be a case where the sample name in the bam's SM field wouldn't match the sample name provided in the YAML's `readgroup` input, which looks like:

```
readgroups:
- "@RG\tID:2900998520\tPU:HNMTJDSXX.1.ACAGCAAC-AGTCAGGA\tSM:H_KA-894912-M1513147_1\tLB:H_KA-894912-M1513147_1-lg1-lib1\tPL:Illumina\tCN:WUGSC"
```

A sanity check to make sure they match (and if not, do the rename_vcf step) might be in order.  Or maybe, just rename the VCF anyway in the mutect subworkflow to ensure that it always matches.

chrisamiller: To clarify, I think there are three options here:
1) always run `replace_vcf_sample_name.cwl` on the vcf files from Mutect, just to make sure things match up (and don't worry about the possible mismatch with the bam header
2) Enforce that the `SM:` tags and sample name inputs match by adding a validation step 
3) Enforce that the `SM:` tags and sample name inputs match by generating or altering the readgroup string in the alignment workflow (using the sample_name provided).

definitions/tools/add_strelka_gt.cwl
johnegarza: Good point, we're not even that far off from `focal` now.. updating

johnegarza:
The latest 3 commits include some changes that should be able to handle most non-alphanumeric/underscore characters. In particular, 7531c3b changes one of the regexes used to extract the sample name from bam files. This is adapted from the general header regex described in [section 1.3 of the SAM specification](https://samtools.github.io/hts-specs/SAMv1.pdf); this new regex should work with any properly formatted bam file, and is consistent with the logic used to extract the sample name elsewhere, such as mutect. While these changes may still not handle every possible case, as discussed at our last meeting, our pre-processing steps will usually set things up correctly. This should also ensure that our assumption that the `SM` tag embedded in input bams will match up with the input sample names.

definitions/tools/mutect.cwl
johnegarza: See my comment below

definitions/tools/mutect.cwl
```diff
@@ -21,15 +21,8 @@ requirements:
             
             TUMOR=`(samtools view -H $3 | grep "^@RG" | sed "s/.*SM:\([^\t]*\).*/\1/g" | head -n 1)` #Extracting the sample name from the TUMOR bam.
             NORMAL=`(samtools view -H $4 | grep "^@RG" | sed "s/.*SM:\([^\t]*\).*/\1/g" | head -n 1)` #Extracting the sample name from the NORMAL bam.
```
chrisamiller: Should this regex be changed to match the one in the filter_docm tool that you used, for consistency?

definitions/tools/replace_vcf_sample_name.cwl
```diff
@@ -18,7 +18,13 @@ requirements:
...
           basen=`basename "$3"`
           basen="renamed.$basen"
-          echo "$1 $2" > sample_update.txt
+
+          #escape spaces, otherwise bcftools will try to use them as a delimiter
+          #triple backslash to escape within backticks and then again within sed
+          old_name=`echo "$1" | sed 's/ /\\\ /g'`
+          new_name=`echo "$2" | sed 's/ /\\\ /g'`
+
+          echo "$old_name $new_name" > sample_update.txt
```
tmooney: Part of me wants to suggest switching to the one-sample per line format, but that'd require this tool to handle all the samples simultaneously.  Got to love custom space-delimited formats!

definitions/tools/mutect.cwl
johnegarza: Agreed, fixed in 29a3f2c

definitions/tools/replace_vcf_sample_name.cwl
johnegarza: For future reference, this is documented [in the bcftools docs](https://samtools.github.io/bcftools/bcftools.html#reheader). I agree that the sample per line format would have been easier to use, but as you mentioned, I'd then also have to handle all samples at once, and figure out what order they're recorded in the vcf to do this properly. This way is a little more of a hack, but safer for our use case.

PR #809, opened 2019-11-12
Add cle disclaimer information to tumor filtered vcf

dufeiyu:
Dave requests to add CLE disclaimer information to tumor filtered vcf.

### DIFF
--- a/definitions/pipelines/aml_trio_cle.cwl
+++ b/definitions/pipelines/aml_trio_cle.cwl
@@ -160,13 +160,8 @@ inputs:
         default: [HGVSc,HGVSp]
     disclaimer_text:
         type: string?
-        default: "#This laboratory developed test (LDT) was developed and its performance characteristics determined by the CLIA Licensed Environment laboratory at the McDonnell Genome Institute at Washington University (MGI-CLE, CLIA #26D2092546, CAP #9047655), Dr. David H. Spencer MD, PhD, FCAP, Medical Director. 4444 Forest Park Avenue, Rm 4127 St. Louis, Missouri 63108 (314) 286-1460 Fax: (314) 286-1810. The MGI-CLE laboratory is regulated under CLIA as certified to perform high-complexity testing. This test has not been cleared or approved by the FDA."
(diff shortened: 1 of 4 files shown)


definitions/pipelines/aml_trio_cle.cwl
```diff
@@ -163,6 +163,11 @@ inputs:
...
     disclaimer_version:
         type: string
+    vcf_disclaimer_text:
+        type: string?
+        default: "##DisclaimerText=This laboratory developed test (LDT) was developed and its performance characteristics determined by the CLIA Licensed Environment laboratory at the McDonnell Genome Institute at Washington University (MGI-CLE, CLIA #26D2092546, CAP #9047655), Dr. David H. Spencer MD, PhD, FCAP, Medical Director. 4444 Forest Park Avenue, Rm 4127 St. Louis, Missouri 63108 (314) 286-1460 Fax: (314) 286-1810. The MGI-CLE laboratory is regulated under CLIA as certified to perform high-complexity testing. This test has not been cleared or approved by the FDA."
```
tmooney: If we wanted to avoid repeating this body of text as two different inputs, this could be done similarly to the disclaimer text in `germline_wgs.cwl`:
https://github.com/genome/analysis-workflows/blob/1585504ccffafac53b1594349ed934f45206ee2b/definitions/pipelines/germline_wgs.cwl#L131-L133
followed by this for VCFs:
https://github.com/genome/analysis-workflows/blob/1585504ccffafac53b1594349ed934f45206ee2b/definitions/pipelines/germline_wgs.cwl#L356-L358
and this for TSVs:
https://github.com/genome/analysis-workflows/blob/1585504ccffafac53b1594349ed934f45206ee2b/definitions/pipelines/germline_wgs.cwl#L396-L398

dufeiyu: Good point. That is a best practice to avoid duplication. Yes, I will make above suggested changes along with output_name (to avoid the annoying xxx.commented.commented.xxx).

PR #811, opened 2019-11-19
Added cle variant filter

johnegarza:
This PR adds a new tool, `filter_cle_variants`. It takes in an input target vcf and a second vcf containing CLE validated variants, and adds a "CLE" tag to the INFO field of variants in the target file that are present in the CLE file. 
This is intended for the immuno workflow, but was added to the `filter_vcf` subworkflow, which is shared by many pipelines. The input vcf is already present in all other calling pipelines, and the CLE vcf is optional. When the CLE vcf is not provided, this step does nothing beyond renaming the input file and adding the line `##FILTER=<ID=PASS,Description="All filters passed">`, so it shouldn't effect the other pipelines. 
Note that the input vcf to this step must have an index file; due to this, I added in a `secondaryFiles` requirement to the `filter_vcf` subworkflow, but since all other pipelines that call this already pass an indexed vcf, this shouldn't change anything.

### DIFF
--- a/definitions/tools/filter_cle_variants.cwl
+++ b/definitions/tools/filter_cle_variants.cwl
@@ -1,0 +1,38 @@
+#!/usr/bin/env cwl-runner
+ 
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Adds an INFO tag (CLE) to variants in the target file present in a CLE-validated vcf file"
+
+requirements:
+    - class: ShellCommandRequirement
+    - class: InlineJavascriptRequirement
+    - class: DockerRequirement
+      dockerPull: "mgibio/bcftools-cwl:1.9"
+    - class: ResourceRequirement
+      ramMin: 8000
+
+baseCommand: ["/opt/bcftools/bin/bcftools", "annotate"]
+arguments:
+    [ "-Oz", "-o", "cle_variants_flagged.vcf.gz" ]  
+
+inputs:
+    cle_variants:
+        type: File?
+        secondaryFiles: [.tbi]
+        inputBinding:
+            position: 1
+            valueFrom: |
+                ${
+                    return [ '-a', inputs.cle_variants.path, '-m', 'CLE' ];
+                }
+    vcf:
+        type: File
+        secondaryFiles: [.tbi]
+        inputBinding:
+            position: 2
+outputs:
+    cle_flagged:
+        type: File
+        outputBinding:
+            glob: "cle_variants_flagged.vcf.gz"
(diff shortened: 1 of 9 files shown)


definitions/tools/filter_cle_variants.cwl
```diff
@@ -0,0 +1,38 @@
...
+
+inputs:
+    cle_variants:
+        type: File?
+        secondaryFiles: [.tbi]
+        inputBinding:
+            position: 1
+            valueFrom: |
+                ${
+                    return [ '-a', inputs.cle_variants.path, '-m', 'CLE' ];
```
tmooney: I think this could be simplified to something like:
```suggestion
                    return [ '-a', self.path, '-m', 'CLE' ];
```

definitions/tools/filter_cle_variants.cwl
```diff
@@ -0,0 +1,38 @@
...
+    vcf:
+        type: File
+        secondaryFiles: [.tbi]
+        inputBinding:
+            position: 2
+outputs:
+    cle_flagged:
+        type: File
+        outputBinding:
+            glob: "cle_variants_flagged.vcf.gz"
```
tmooney: Do we care about being able to customize this name or base it on the input name?

definitions/tools/filter_cle_variants.cwl
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env cwl-runner
+ 
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Adds an INFO tag (CLE) to variants in the target file present in a CLE-validated vcf file"
+
```
chrisamiller: Could we make this name a bit more general, since I can envision other uses for it (looking for specific variants in recurrence samples, etc)?  Maybe call it "filter_existing_variants.cwl" and instead of labelling with the "CLE" tag, use "EXISTING" (or something comparable).    Update: Jason suggests "previously_discovered"

definitions/tools/filter_cle_variants.cwl
johnegarza: Accepted in 8b52093

definitions/tools/filter_cle_variants.cwl
johnegarza: Done in 8b52093

definitions/tools/filter_cle_variants.cwl
johnegarza: Updated this to be more general in the latest commit. Since this output isn't exposed anywhere except to a downstream tool that immediately renames it, I think it's ok as-is.

tmooney: Definitely works!  In filters I'm currently biased towards basing the output name on the input name by default, just because that's a great way to disambiguate if the same step has to run multiple times in a workflow (especially if both end up being outputs).  None of that currently applies in this case, though! :smile:

PR #824, opened 2019-12-03
use a docker image tag for rnaseq tools

apaul7:
removes the latest image tags for the cwl tools. future images and PRs should be made to use smaller images instead of the large rnaseq one. 

WIP until the docker image creation succeeds on docker hub


apaul7:
the rna seq docker image `1.0.0` was created and this pr is ready for review

PR #825, opened 2019-12-03
Fixing warnings: subworkflow directory low-hanging fruit edition

tmooney:
This fixes some validator warnings in our subworkflows that result from one of a few circumstances:
* an optional input with default is mapped to a required input:
  this is made required to eliminate the ability to pass `null` overriding the default
* some missing `secondaryFile` declarations have been added.


tmooney:
Also note that some remaining warnings are of the type described here: common-workflow-language/cwltool#759

PR #828, opened 2019-12-03
Add pyyaml dependency

acoffman:
also remove debug printout


.travis.yml
```diff
@@ -6,6 +6,8 @@ install:
     - pip install 'cwltool==1.0.20181217162649'
     - pip install 'ruamel.yaml==0.15.77'
     - pip install 'mdutils==1.0.0'
+    - pip install 'mdutils==1.0.0'
```
susannasiebert: This seems to be an accidental duplication.

acoffman: Good catch, fixed

PR #840, opened 2019-12-09
Updated UMI workflow for basic UMI-aware alignment and qc

jasonwalker80:
first pass at saving all BAMs/CRAMs that might be needed


chrisamiller:
Suggestion to add some "doc:" lines that explain the difference between the output crams

PR #848, opened 2019-12-26
A downsample and re-call pipeline.

tmooney:
For taking existing CRAMs and re-running HaplotypeCaller on downsampled versions thereof.


tmooney:
Yeah, I hope that use of this particular custom type doesn't proliferate beyond this pipeline (and its "gathered" version), but if it does it would definitely make sense to extract it to a custom type.

PR #851, opened 2020-01-08
Minimal changes to ensure unique immuno output names

johnegarza:
Previously, there were 2 pairs of output files with the same name.
`annotated.vcf_summary.html` was a hardcoded name in vep, run in both the `somatic_exome` and `germline_exome` pipeline. This now pulls part of its name from the input vcf.
`variants.annotated.tsv` had an optional unused prefix in add_vep_fields_to_table, run in both `somatic_exome` and `pvacseq`. I added a new input with a default that adds a prefix to the pvacseq branch. Neither change requires any new inputs.

### DIFF
--- a/definitions/pipelines/immuno.cwl
+++ b/definitions/pipelines/immuno.cwl
@@ -285,6 +285,9 @@ inputs:
         type: boolean?
     pvacseq_threads:
         type: int?
+    vep_to_table_prefix:
+        type: string?
+        default: 'pvacseq'
 
     tumor_sample_name:
         type: string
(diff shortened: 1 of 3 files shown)


definitions/pipelines/immuno.cwl
```diff
@@ -285,6 +285,9 @@ inputs:
         type: boolean?
     pvacseq_threads:
         type: int?
+    vep_to_table_prefix:
+        type: string?
+        default: 'pvacseq'
```
tmooney: Should this default be in the `pvacseq.cwl`?

johnegarza: I was trying to minimize changes outside of the immuno.cwl, but this change can probably go there instead without disrupting anything else

definitions/subworkflows/pvacseq.cwl
```diff
@@ -106,6 +106,9 @@ inputs:
     vep_to_table_fields:
         type: string[]?
         default: [HGVSc,HGVSp]
+    vep_to_table_prefix:
```
susannasiebert: Do we actually need this to be an input here since we are not using it in the immuno.cwl? I believe you can just defined this default in the vep.cwl step directly.

PR #860, opened 2020-02-04
Added doc line

bryanfisk:
Added doc line for manta_non_wgs input.


chrisamiller:
might be easier to make a branch, save up a bunch of these, and then make a large pull request, rather than having one PR for each line that's being changed!

PR #866, opened 2020-02-04
Immuno cwl docs - Somatic inputs

jhundal:


definitions/pipelines/immuno.cwl
```diff
@@ -59,32 +59,103 @@ inputs:
...
+          Appropriate files can be found on ensembl at https://ensembl.org/info/data/ftp/index.html
+          When providing the reference secondary files corresponding to reference indices must be
+          located in the same directory as the reference itself. These files can be created with
+          samtools index, bwa index, and picard CreateSequenceDictionary.
 
     tumor_sequence:
         type: ../types/sequence_data.yml#sequence_data[]
+        label: "tumor_sequence: yml file specifying the location of MT sequencing data"
+        doc: |
+          tumor_sequence is a yml file for which to pass information regarding
```
tmooney: I don't think it's quite right to say that this is a `yml file`.  It's a data structure that happens to be described in `sequence_data.yml`, but the data structure itself needn't be provided in its own file.  It's part of the regular inputs.

definitions/pipelines/immuno.cwl
```diff
@@ -59,32 +59,103 @@ inputs:
...
+          for the entire chromosome (i.e. chr1, chr2, etc.), these names should match
+          the format in the reference file.
     bait_intervals:
         type: File
+        label: "bait_intervals: interval_list file of baits used in the sequencing experiment"
+        doc: |
+          bait_intervals is an interval_list corresponding to the baits used in sequencing reagent.
+          These are essentially coordinates for regions you were able to design probes for in the reagent.
+          Typically the reagent provider has this information available in bed format and it can be
+          converted to an interval_list with Picards BedToIntervalList. Astrazeneca also maintains a repo
```
tmooney:
```suggestion
          converted to an interval_list with Picards BedToIntervalList. AstraZeneca also maintains a repo
```

definitions/pipelines/immuno.cwl
jasonwalker80: I might suggest also modifying (or working with @zlskidmore or @bryanfisk) the somatic_exome.cwl doc/label for these same inputs:
https://github.com/genome/analysis-workflows/blob/cfdcd0735e8fb1ac0b97f1a5cbe0cc7e00033dd0/definitions/pipelines/somatic_exome.cwl#L39

PR #882, opened 2020-02-04
Immuo CWL output doc update2 for somatic_exome/exome_alignment_and_qc

gschang:
I updated the rnaseq subworkflow for documentation of the output fields.

### DIFF
--- a/definitions/pipelines/immuno.cwl
+++ b/definitions/pipelines/immuno.cwl
@@ -347,81 +347,153 @@ outputs:
     tumor_cram:
         type: File
         outputSource: somatic/tumor_cram
+        label: "Sorted CRAM from tumor DNA"
+        doc: |
+          Sorted CRAM file of sequencing read alignments by bwa-mem from a tumor DNA sample with duplicate reads tagged
     tumor_mark_duplicates_metrics:
         type: File
         outputSource: somatic/tumor_mark_duplicates_metrics
+        label: "Sequencing duplicate metrics from tumor DNA"
+        doc: |
+          Duplication metrics on duplicate sequencing reads from a tumor DNA sample, identified by picard MarkDuplicates tool
     tumor_insert_size_metrics:
         type: File
         outputSource: somatic/tumor_insert_size_metrics
+        label: "Paired-end sequencing diagnostics/quality metrics from tumor DNA"
+        doc: |
+          Diagnostics/quality metrics including the insert size distribution and read orientation of the paired-end libraries from a tumor DNA sample
     tumor_alignment_summary_metrics:
         type: File
         outputSource: somatic/tumor_alignment_summary_metrics
+        label: "Sequencign alignment summary from tumor DNA"
+        doc: |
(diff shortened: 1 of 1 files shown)


gschang:
I have updated the doc in the immuno workflow. The first revision about rnaseq has been approved. Next, this new revision is the second one, and about the exome alignment and QC.

definitions/pipelines/immuno.cwl
```diff
@@ -298,106 +298,202 @@ outputs:
...
         outputSource: rnaseq/transcript_abundance_h5
+        label: "Transcript-level abundance table in HDF5 format by kallisto"
+        doc: |
+          HDF5 binary file containing transcript-level abundance esimates, bootstrap estimate, and so on, created by kallisto
     gene_abundance:
         type: File
         outputSource: rnaseq/gene_abundance
+        label: "Gene-level abundance output by tximport with kallisto output"
+        doc: |
+          Tab-delmited file containing the abundance estimates summarized in the gene level with kallisto output by Bioconductor tximport tool
```
tmooney:
```suggestion
          Tab-delimited file containing the abundance estimates summarized in the gene level with kallisto output by Bioconductor tximport tool
```

gschang: I will fix this mistake. Thank you!

definitions/pipelines/immuno.cwl
```diff
@@ -298,106 +298,202 @@ outputs:
...
         outputSource: somatic/tumor_hs_metrics
+        label: "Sequencing coverage summary of target intervals from tumor DNA"
+        doc: |
+          Diagnostics/quality metrics specific for sequencing data generated through hybrid-selection (e.g. whole exome) from a tumor DNA sample, for example to assess target coverage
     tumor_per_target_coverage_metrics:
         type: File[]
         outputSource: somatic/tumor_per_target_coverage_metrics
+        label: "Sequencing per-target coverage summary of 59 ACMG genes from tumor DNA"
+        doc: |
+          Diagnostics/quality metrics showing detailed sequencing coverage per target interval for 59 genes (source: ACMG) from a tumor DNA sample
```
tmooney: Technically these references to ACMG and ClinVar are not specifically true of this pipeline.  It's entirely a matter of what inputs someone chooses to provide.

gschang: I agree that this parameters don't specify which input. Actual pipeline run (i.e. inputs) was intended for these specific targets--(1) ClinVar variants and (2) ACMG genes. In fact, the parameter names themselves in CWL sound confusing in this discrepancy. Maybe, "tumor_per_target_coverage_acmg_metrics" may sound more reasonable in the workflow. However, I am focusing doc/label in this revision.

tmooney: The documentation might suggest that certain inputs would be useful, but I think it is too specific to suggest they are the only possibilities.  For example, some other gene list might be useful for a different project that uses this same pipeline.  The fact that one project wants to use 59 particular genes is more a character of that project than of the pipeline itself.

gschang: I got your point. I will update this doc, for more general purpose.

gschang:
As Tom suggested (above), I updated several fields for general purpose. I also made a correction in some misspelling.

PR #884, opened 2020-02-06
Better names for HLA allele parameters

johnegarza:


jhundal:
How about clinical_hla and optitype_hla?

gschang:
In fact, the pvacseq CWL use this parameter as "alleles". "optitype_alleles" and "clinical_alleles" look consistent for the current parameter in pvacseq.

Issue #837, opened 2019-12-06
smoove.cwl error: "Failed to open -: unknown file type"


tatahin:
Good day!
I keep trying to run `germline_wgs.cwl` with BAMs and references from `example_data/exome_workflow` from repository (I had to add some files and values). But it failed with error:
```
 [job run_smoove] Skipping Docker software container '--memory' limit despite presence of ResourceRequirement with ramMin and/or ramMax setting. Consider running with --strict-memory-limit for increased portability assurance.
INFO [job run_smoove] /home/bio/serge/analysis-workflows_TATAHIN_FORK/results/05122019_3/Temp/l3aef3eb$ docker \
    run \
    -i \
    --volume=/home/bio/serge/analysis-workflows_TATAHIN_FORK/results/05122019_3/Temp/l3aef3eb:/RJsOtd:rw \
    --volume=/tmp/qcvuiv0a:/tmp:rw \
    --volume=/home/bio/serge/analysis-workflows_TATAHIN_FORK/results/05122019_3/Temp/27qxdb0a/final.bam:/var/lib/cwl/stga84ed033-02ba-4316-9104-711b03107983/final.bam:ro \
    --volume=/home/bio/serge/analysis-workflows_TATAHIN_FORK/results/05122019_3/Temp/5qcp_6ca/final.bam.bai:/var/lib/cwl/stga84ed033-02ba-4316-9104-711b03107983/final.bam.bai:ro \
    --volume=/home/bio/serge/analysis-workflows_TATAHIN_FORK/results/05122019_3/Temp/5qcp_6ca/final.bai:/var/lib/cwl/stga84ed033-02ba-4316-9104-711b03107983/final.bai:ro \
    --volume=/home/bio/serge/analysis-workflows_TATAHIN_FORK/example_data/exome_workflow/chr17_test.fa:/var/lib/cwl/stg8e064cc4-5d8a-48e7-ba03-9ebda8c10492/chr17_test.fa:ro \
    --volume=/home/bio/serge/analysis-workflows_TATAHIN_FORK/example_data/exome_workflow/chr17_test.fa.fai:/var/lib/cwl/stg8e064cc4-5d8a-48e7-ba03-9ebda8c10492/chr17_test.fa.fai:ro \
[...]

Issue #726, opened 2019-07-10
Copy number reference for CNVkit


Stikus:
Hello. I'm trying to test your [germline_wgs.cwl](https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/germline_wgs.cwl) pipeline with [germline_wgs.yaml](https://github.com/genome/analysis-workflows/blob/master/example_data/germline_wgs.yaml) as values file.
I have several problems:
- Looks like [this](https://github.com/genome/analysis-workflows/blob/master/definitions/subworkflows/cnvkit_single_sample.cwl#L81) line is wrong - [cnvkit_vcf_export.cwl](https://github.com/genome/analysis-workflows/blob/master/definitions/tools/cnvkit_vcf_export.cwl) doesn't have `tumor_bam` input. 
- According to these lines: [germline_wgs.cwl#L88](https://github.com/genome/analysis-workflows/blob/master/definitions/pipelines/germline_wgs.cwl#L88), [single_sample_sv_callers.cwl#L29](https://github.com/genome/analysis-workflows/blob/master/definitions/subworkflows/single_sample_sv_callers.cwl#L29) and [cnvkit_single_sample.cwl#L24](https://github.com/genome/analysis-workflows/blob/master/definitions/subworkflows/cnvkit_single_sample.cwl#L24) - `cnvkit_reference_cnn` is required input.
But according to [cnvkit_batch.cwl#L67](https://github.com/genome/analysis-workflows/blob/master/definitions/tools/cnvkit_batch.cwl#L67) and tool help - it is optional.
I ask this question because [exome_workflow.tgz](https://github.com/genome/analysis-workflows/blob/master/example_data/exome_workflow.tgz) doesn't contain this file and [germline_wgs.yaml](https://github.com/genome/analysis-workflows/blob/master/example_data/germline_wgs.yaml) doesn't contain this input.

So - is this reference required? And if yes - how should I get it?
I find [this](https://cnvkit.readthedocs.io/en/stable/pipeline.html#reference) method - but I doesn't have these files either.

tirohia:
Not sure whether this should go here, or a separate ticket opened. 

I had the same issue, and using cnvkit, pooling all available normal samples to get a reference.cnn file,

```cnvkit.py batch *.tumor.bam -n *.normal.bam -m wgs -f /blue/project/dev/data/references/hg38/hg38.fa --output-reference reference.cnn
```

 placing it in the analysis-workflows/example_data/ directory and adding the line:
[...]

PR #885, opened 2020-02-18
CNVkit and other SV updates

apaul7:
This PR has multiple updates:
* updates cnvkit to default to flat reference
  this allows the SV callers to run without providing a `reference.cnn` input (#726)
* adds cnvkit input `target_average_size `
  I have not linked this in with the pipeline. 
* makes the cnvkit `method` input an enum
* remove default method for cnvkit single sample
  the cnvkit batch tool already has the `hybrid` default
* make `wgs` default method for cnvkit in wgs pipeline
* add some cnvkit input doc lines at the tool level
* update AnnotSV tool to use full path
* update some reference inputs in sv tools to allow `File` or `string`
* updated example `germline_wgs.yaml`
  the exome bams used previously cannot succeed in Smoove (#837)
  the somatic wgs example yaml references `/gscmnt/gc2560/core/cwl/example_data/HCC1395_normal_downsampled_wgs.bam` 
  The workflow is able to succeed using that bam and full GRCh38 reference. Should it be added to the repo or hosted elsewhere?

### DIFF
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -23,7 +23,7 @@ inputs:
     male_reference:
         type: boolean?
     reference_cnn:
-        type: File
+        type: File?
     cnvkit_vcf_name:
         type: string
         default: "cnvkit.vcf"
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -76,7 +76,7 @@ steps:
                     ${
                       var cnn = self[0];
                       var fasta = self[1];
-                      return {'reference': {'reference_cnn': cnn, 'reference_fasta': fasta}};
+                      return {'cnn_file': cnn, 'fasta_file': fasta};
                     }
         out:
             [cn_diagram, cn_scatter_plot, tumor_antitarget_coverage, tumor_target_coverage, tumor_bin_level_ratios, tumor_segmented_ratios]            
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -70,8 +70,14 @@ steps:
             scatter_plot: scatter_plot
             drop_low_coverage: drop_low_coverage
             male_reference: male_reference
-            reference_cnn: reference_cnn
-            reference: fasta_reference
+            reference:
+                source: [reference_cnn, fasta_reference]
+                valueFrom: |
+                    ${
+                      var cnn = self[0];
+                      var fasta = self[1];
+                      return {'reference': {'reference_cnn': cnn, 'reference_fasta': fasta}};
+                    }
         out:
             [cn_diagram, cn_scatter_plot, tumor_antitarget_coverage, tumor_target_coverage, tumor_bin_level_ratios, tumor_segmented_ratios]            
             
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -10,8 +10,10 @@ inputs:
     tumor_bam:
         type: File
     method:
-        type: string?
-        default: "hybrid"
+        type:
+          - "null"
+          - type: enum
+            symbols: ["hybrid", "amplicon", "wgs"]
     diagram:
         type: boolean?
     scatter_plot:
--- a/definitions/subworkflows/cnvkit_single_sample.cwl
+++ b/definitions/subworkflows/cnvkit_single_sample.cwl
@@ -7,6 +7,11 @@ requirements:
     - class: StepInputExpressionRequirement
 
 inputs:
+    fasta_reference:
+        type:
+            - string
+            - File
+        secondaryFiles: [.fai]
     tumor_bam:
         type: File
     method:
@@ -66,6 +71,7 @@ steps:
             drop_low_coverage: drop_low_coverage
             male_reference: male_reference
             reference_cnn: reference_cnn
+            reference: fasta_reference
         out:
             [cn_diagram, cn_scatter_plot, tumor_antitarget_coverage, tumor_target_coverage, tumor_bin_level_ratios, tumor_segmented_ratios]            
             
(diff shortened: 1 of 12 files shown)


definitions/tools/cnvkit_batch.cwl
```diff
@@ -13,16 +13,19 @@ requirements:
       ramMin: 4000
       tmpdirMin: 10000
     - class: InlineJavascriptRequirement
+arguments:
+    ["--normal"]
```
tmooney: Does this work to have this repeated here and for the `normal_bam` input?  The help text makes it sound like it wants to be followed by 0 or more files, but it looks a little strange to specify it twice.

apaul7: I think it's okay to have `--normal` `other inputs` `--normal actual_normal.bam`. I will double check. 
I'm sure there is a way to force the `--normal` flag to be passed even if no normal bam input is provided. Adding it to the arguments was the easiest for me.

tmooney: One way is to keep it in arguments like you have and then make sure the `normal_bam` input is first after the arguments (with no specified prefix).  I think right now you'll end up with `--normal --normal x.bam` as it is written.

apaul7:
one of my tests failed on cnvkit when I provide the `reference.cnn`. Working on modifying the cnvkit to default to flat reference unless the `reference.cnn` is provided.
```
If -r/--reference is given, options to construct a new reference (-n/--normal, -f/--fasta) should not be used.
```

definitions/tools/cnvkit_batch.cwl
apaul7: ahh what a good solution!

apaul7:
This is ready for review now. 

The cnvkit_batch.cwl now has a reference record input. This allows the tool to run in a few different modes. Defaulting to the reference.cnn if it's provided. 
* flat reference
  provide just the `reference.fasta_file`
* tumor/normal
  provide the `reference.fasta_file` and `reference.normal_bam`
* reference.cnn
  provide the `reference.cnn_file` input

```arguments: [{ valueFrom: "$((inputs.reference.hasOwnProperty('cnn_file'))? '' : '--normal')" }]```
is used to add `--normal` when the cnn_file is not provided as cnvkit complains if it's present when running in the reference.cnn mode.

File: definitions/tools/cnvkit_batch.cwl [hunk 2]
```diff
@@ -13,64 +13,89 @@ requirements:
       ramMin: 4000
       tmpdirMin: 10000
     - class: InlineJavascriptRequirement
+arguments: [{ valueFrom: "$((inputs.reference.hasOwnProperty('cnn_file'))? '' : '--normal')" }]
```
tmooney: This seems like it might yield an extra positional argument that is the empty string.  Maybe that's harmless.  If it is problematic, you could possibly convert this to a boolean with a prefix.  Then it should follow the normal CWL rules of only including the prefix on `true` values.

definitions/tools/cnvkit_batch.cwl
```diff
@@ -13,64 +13,89 @@ requirements:
...
+              type:
+                - string
+                - File
+              inputBinding:
+                position: 2
+                prefix: "--fasta"
+            normal_bam:
+              type: File?
+              inputBinding:
+                position: 1
```
tmooney: Alternatively, I wonder if you could get away with defining a `valueFrom` here that always includes the `--normal` and optionally includes the file when it's present.

definitions/subworkflows/cnvkit_single_sample.cwl
```diff
@@ -63,7 +70,14 @@ steps:
...
             drop_low_coverage: drop_low_coverage
             male_reference: male_reference
-            reference_cnn: reference_cnn
+            reference:
+                source: [reference_cnn, fasta_reference]
+                valueFrom: |
+                    ${
+                      var cnn = self[0];
+                      var fasta = self[1];
+                      return {'cnn_file': cnn, 'fasta_file': fasta};
```
tmooney: This does not match either of the types available for this input.

tmooney: The reference input of `cnvkit_batch.cwl` is defined as one of the following:
* A `record` containing a `cnn_file`.
* A `record` containing a `fasta_file` and an optional `normal_bam`.

Just like `fasta_file` oughn't to be given both a `string` and `File` simultaneously, the type definition is satisfied by either one of these two records.  The record with `cnn_file` has no `fasta_file` and the record with `fasta_file` has no `cnn_file`.

apaul7: I'm using the exclusive input parameter as described in https://www.commonwl.org/user_guide/11-records/index.html
if both records are provided it will take the first record type. In this case that's the record with the `cnn_file`

tmooney: After a quick test I concede that, although `cwltool` produces an error message in this situation (like:
```
echo12.yaml:3:5: invalid field `thing2`, expected one of: 'thing1'
```
), it seems to carry on regardless.

That being the case I might prefer to specifically give it one or the other parameter here while we're already scripting, but it seems to not be strictly necessary.

apaul7: Are you suggesting to add more javascript to check if the `cnn_file` is non null? 
if it is, pass just `cnn_file: cnn`
if it is not, pass just `fasta_file: fasta`

or are you suggesting basically 2 cnvkit workflows, 
1 for `reference.cnn` method
1 for the flat reference and tumor/normal methods 

Trying to make the workflows easy to use and configure but don't want to make it super complicated!

tmooney: I was imagining the "more javascript" route.  The complexity remains hidden inside the code that way.  If it works fine in Cromwell, though, maybe changing it isn't necessary at all.

apaul7: I'm hesitant to add more javascript. I'm bad at debugging javascript in the cwl files and the errors Cromwell displays are not useful. 
If you have a solution I can add it and rerun my tests.

PR #887, opened 2020-02-27
ensure unique output names to prevent gms staging failures

johnegarza:
Current GMS immuno builds are failing because pvacseq outputs 3 sets of 3 files, where each set contains files with identical names. Pvacseq outputs these into 3 subdirectories, but the cwl extracts each file individually, causing naming collisions. Renaming the files could work, as described in the "Rename an output file" section [here](https://www.commonwl.org/user_guide/misc/); however, Cromwell does not support this option. In order to avoid sending each of these 9 files to a tool just to rename, I opted to instead just grab each of the 3 output directories. They do include some additional temp and log files, but they aren't very large and can easily be cleaned up later.

### DIFF
--- a/definitions/tools/pvacseq.cwl
+++ b/definitions/tools/pvacseq.cwl
@@ -16,7 +16,7 @@ arguments: [
     "/opt/conda/bin/pvacseq", "run",
     "--iedb-install-directory", "/opt/iedb",
     "--pass-only",
-    { position: 5, valueFrom: "pvacseq_epitopes" },
+    { position: 5, valueFrom: "pvacseq_predictions" },
 ]
 requirements:
     - class: ShellCommandRequirement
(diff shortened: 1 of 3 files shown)


johnegarza:
For future reference: there is an open bug report concerning the inability to rename [here](https://broadworkbench.atlassian.net/browse/BA-5835). The original workaround in this PR failed due to another bug (Cromwell can't handle type `Directory?`- bug report to be filed soon). The latest commit adds in another workaround by creating a new output directory containing all of the results. Since this directory will always be there, the output can be type `Directory`, which Cromwell handles just fine.

definitions/tools/pvacseq.cwl
```diff
@@ -16,7 +16,7 @@ arguments: [
     "/opt/conda/bin/pvacseq", "run",
     "--iedb-install-directory", "/opt/iedb",
     "--pass-only",
-    { position: 5, valueFrom: $(runtime.outdir) },
+    { position: 5, valueFrom: "pvacseq_epitopes" },
```
susannasiebert: I'm not super familiar with how Cromwell resolves cases like this. Where would it create this directory? Inside of the process's working directory?

johnegarza: Yup, within the `execution` directory for this step

PR #891, opened 2020-03-16
Better handling of MHC Class II alleles

johnegarza:
MHC class I alleles follow a standard nomenclature, but not all class II alleles follow it. This, in addition to the notation used for uncertain clinical calls and paired class II alleles, makes parsing a single input difficult. Breaking up the input into class I and class II allows most of the current parsing script to remain usable, while eliminating most of the edge cases caused by class II alleles.

### DIFF
--- a/definitions/pipelines/immuno.cwl
+++ b/definitions/pipelines/immuno.cwl
@@ -288,11 +288,14 @@ inputs:
     reference_dict:
         type: File
 
-    clinical_hla_alleles:
+    clinical_MHCcI_alleles:
         type: string[]?
-        label: "clinical_calls: Clinical HLA typing results; element format: HLA-X*01:02[/HLA-X...]"
-        doc: |
-          clinical_calls is used to provide clinical HLA typing results in the format HLA-X*01:02[/HLA-X...] when available.
+        label: "Clinical HLA typing results, limited to MHC Class I alleles; element format: HLA-X*01:02[/HLA-X...]"
+        doc: "used to provide clinical HLA typing results in the format HLA-X*01:02[/HLA-X...] when available."
+    clinical_MHCcII_alleles:
+        type: string[]?
+        label: "Clinical HLA typing results, limited to MHC Class II alleles"
+        doc: "clinical_calls is used to provide clinical HLA typing results; separated from class I due to nomenclature inconsistencies"
 
     #pvacseq inputs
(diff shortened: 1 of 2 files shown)


tmooney:
Names similar to what is seen in the pvacseq tool could work, too. E.g.:
https://github.com/genome/analysis-workflows/blob/ff99dc68aff782c6b95781b774db4e6f5b3bd907/definitions/tools/pvacseq.cwl#L170
https://github.com/genome/analysis-workflows/blob/ff99dc68aff782c6b95781b774db4e6f5b3bd907/definitions/tools/pvacseq.cwl#L182

definitions/pipelines/immuno.cwl
```diff
@@ -288,11 +288,14 @@ inputs:
...
         type: string[]?
-        label: "clinical_calls: Clinical HLA typing results; element format: HLA-X*01:02[/HLA-X...]"
-        doc: |
-          clinical_calls is used to provide clinical HLA typing results in the format HLA-X*01:02[/HLA-X...] when available.
+        label: "Clinical HLA typing results, limited to MHC Class I alleles; element format: HLA-X*01:02[/HLA-X...]"
+        doc: "used to provide clinical HLA typing results in the format HLA-X*01:02[/HLA-X...] when available."
+    clinical_MHCcII_alleles:
+        type: string[]?
+        label: "Clinical HLA typing results, limited to MHC Class II alleles"
+        doc: "clinical_calls is used to provide clinical HLA typing results; separated from class I due to nomenclature inconsistencies"
```
tmooney: I'm guessing this used to be called `clinical_calls` once upon a time? Could probably just take that part out.
```suggestion
        doc: "used to provide clinical HLA typing results; separated from class I due to nomenclature inconsistencies"
```

definitions/tools/hla_consensus.cwl
```diff
@@ -237,7 +239,13 @@ inputs:
...
+    clinical_MHCcI_alleles:
+        type: string[]?
+        inputBinding:
+            position: 2
+            itemSeparator: ','
+            separate: false
+    clinical_MHCcII_alleles:
         type: string[]?
         inputBinding:
             position: 2
```
tmooney: May want to make sure this position is unique from the class one position, since we're relying on positional parameters in the script.

PR #900, opened 2020-06-05
Add Somatic Concordance Graph

dufeiyu:
There is a need to make plot/graph of somatic concordance cwl outputs. This PR is to add the step of plot_output to the workflow. 

Refer to https://jira.ris.wustl.edu/browse/CI-481

### DIFF
--- a/definitions/tools/somatic_concordance_graph.cwl
+++ b/definitions/tools/somatic_concordance_graph.cwl
@@ -1,0 +1,74 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Generate somatic concordance report graph"
+baseCommand: ["/usr/local/bin/Rscript", "somatic_concordance_graph.R"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "registry.gsc.wustl.edu/mgi-cle/somatic-concordance-report:v1"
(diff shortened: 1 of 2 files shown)


definitions/tools/somatic_concordance_graph.cwl
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Generate somatic concordance report graph"
+baseCommand: ["/usr/local/bin/Rscript", "somatic_concordance_graph.R"]
+requirements:
+    - class: DockerRequirement
+      dockerPull: "registry.gsc.wustl.edu/mgi-cle/somatic-concordance-report:v1"
```
tmooney, line 9: Is there anything in this image that prevents it from being released publicly?

dufeiyu: Not really. This local registry has several pre-installed R libraries and the R script so we can manually run the script. This is very specific for cle group purpose and probably is not much useful outside of cle.

tmooney: Yeah, I don't know who might want to use this tool or workflow.  In the absence of reasons to do otherwise [like licensing restrictions or protected information], I have a bias towards keeping the contents of this repo functional for external users.

PR #908, opened 2020-07-14
Add Flexbar to Sequence Alignment Step

tmooney:
In this version only two options are exposed.  I suspect it'd be good to support all the options seen in the RNA version of trimming (and possibly to harmonize the two approaches to passing trimming parameters!) in order to add this to `immuno.cwl` as well:
https://github.com/genome/analysis-workflows/blob/700e73aaed6db1ad538dd27b2e1709f436ad3edb/definitions/pipelines/immuno.cwl#L32-L41
I thought it better to introduce this first and then tackle that as a separate step.

This is part of genome/covid-19#8 and should support the options currently required by that issue.

### DIFF
--- a/definitions/tools/sequence_align_and_tag.cwl
+++ b/definitions/tools/sequence_align_and_tag.cwl
@@ -9,11 +9,13 @@ requirements:
     - class: SchemaDefRequirement
       types:
           - $import: ../types/sequence_data.yml
+          - $import: ../types/trimming_options.yml
     - class: ResourceRequirement
       coresMin: 8
       ramMin: 20000
+    - class: InlineJavascriptRequirement
     - class: DockerRequirement
-      dockerPull: "mgibio/alignment_helper-cwl:1.0.0"
+      dockerPull: "mgibio/alignment_helper-cwl:1.1.0"
     - class: InitialWorkDirRequirement
       listing:
       - entryname: 'sequence_alignment_helper.sh'
@@ -22,7 +24,7 @@ requirements:
             set -o errexit
             set -o nounset
 
-            while getopts "b:?1:?2:?g:r:n:" opt; do
+            while getopts "b:?1:?2:?g:r:n:t:?o:?" opt; do
                 case "$opt" in
                     b)
                         MODE=bam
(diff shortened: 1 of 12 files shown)


definitions/tools/sequence_align_and_tag.cwl
```diff
@@ -9,11 +9,13 @@ requirements:
...
       types:
           - $import: ../types/sequence_data.yml
+          - $import: ../types/trimming_options.yml
     - class: ResourceRequirement
       coresMin: 8
       ramMin: 20000
+    - class: InlineJavascriptRequirement
     - class: DockerRequirement
-      dockerPull: "mgibio/alignment_helper-cwl:1.0.0"
+      dockerPull: "mgibio/alignment_helper-cwl:1.1.0"
```
dufeiyu, line 18: Assume this version of image will have latest flexbar installed and a PR will be created for it.

definitions/tools/sequence_align_and_tag.cwl
```diff
@@ -45,14 +47,31 @@ requirements:
...
+                        ;;
                 esac
             done
 
             if [[ "$MODE" == 'fastq' ]]; then
-                /usr/local/bin/bwa mem -K 100000000 -t "$NTHREADS" -Y -R "$READGROUP" "$REFERENCE" "$FASTQ1" "$FASTQ2" | /usr/local/bin/samblaster -a --addMateTags | /opt/samtools/bin/samtools view -b -S /dev/stdin
+                if [[ -z \${TRIMMING_ADAPTERS-} || -z \${TRIMMING_ADAPTER_MIN_OVERLAP-} ]]; then
+                    /usr/local/bin/bwa mem -K 100000000 -t "$NTHREADS" -Y -R "$READGROUP" "$REFERENCE" "$FASTQ1" "$FASTQ2" | /usr/local/bin/samblaster -a --addMateTags | /opt/samtools/bin/samtools view -b -S /dev/stdin
+                else
+                    /opt/flexbar/flexbar --adapters "$TRIMMING_ADAPTERS" --reads "FASTQ1" --reads2 "$FASTQ2" --adapter-trim-end LTAIL --adapter-min-overlap "$TRIMMING_ADAPTER_MIN_OVERLAP" --adapter-error-rate 0.1 --max-uncalled 300 --stdout-reads \
```
dufeiyu, line 63: Just wanna make sure two hardcoded parameters "--adapter-error-rate 0.1" and "--max-uncalled 300" are the optimized ones for most cases.

definitions/tools/sequence_align_and_tag.cwl
tmooney: It was added in genome/docker-alignment_helper-cwl#1.  I created the new release branch from master after that PR was merged.

dufeiyu: 👍  You are right. I didn't check it.

PR #916, opened 2020-08-04
Tool to expand interval lists

chrisamiller:
does what it says on the tin

### DIFF
--- a/definitions/tools/interval_list_expand.cwl
+++ b/definitions/tools/interval_list_expand.cwl
@@ -4,13 +4,13 @@ cwlVersion: v1.0
 class: CommandLineTool
 label: "expand interval list regions by a given number of basepairs"
 
-baseCommand: ["/usr/bin/java", "-Xmx16g", "-jar", "/usr/picard/picard.jar", "IntervalListTools"]
+baseCommand: ["/usr/bin/java", "-Xmx3g", "-jar", "/usr/picard/picard.jar", "IntervalListTools"]
 
 arguments:
-    ["OUTPUT=", { valueFrom: $(runtime.outdir)/expanded.interval_list }, "UNIQUE=TRUE"]
+    [valueFrom: "OUTPUT=$(runtime.outdir)/$(inputs.interval_list.nameroot).expanded.interval_list", "UNIQUE=TRUE"]
 requirements:
     - class: ResourceRequirement
-      ramMin: 16000
+      ramMin: 4000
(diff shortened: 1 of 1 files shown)


definitions/tools/interval_list_expand.cwl
```diff
@@ -0,0 +1,29 @@
...
+requirements:
+    - class: ResourceRequirement
+      ramMin: 16000
+    - class: DockerRequirement
+      dockerPull: "mgibio/picard-cwl:2.18.1"
+inputs:
+    interval_list:
+        type: File
+        inputBinding:
+            prefix: "INPUT="
```
tmooney, line 20: Should we go with `separate: false` on these inputs?  (I think Picard is up to the task of dealing with it either way, but that'd be the expected way to do it.)  Alternatively we could upgrade to the GATK4 version, which lets us specify these a different way, like `--INPUT`.

definitions/tools/interval_list_expand.cwl
```diff
@@ -0,0 +1,29 @@
...
+    roi_padding:
+        type: int
+        inputBinding:
+            prefix: "PADDING="
+outputs:
+    expanded_interval_list:
+        type: File
+        outputBinding:
+            glob: "expanded.interval_list"
```
tmooney, line 29: Should we derive this name from the input name?  (I generally prefer it in case someone wants to use this tool multiple times in a given workflow, but it may not be strictly necessary at the moment.)

definitions/tools/interval_list_expand.cwl
```diff
@@ -0,0 +1,29 @@
...
+class: CommandLineTool
+label: "expand interval list regions by a given number of basepairs"
+
+baseCommand: ["/usr/bin/java", "-Xmx16g", "-jar", "/usr/picard/picard.jar", "IntervalListTools"]
+
+arguments:
+    ["OUTPUT=", { valueFrom: $(runtime.outdir)/expanded.interval_list }, "UNIQUE=TRUE"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 16000
```
tmooney, line 13: It's good to leave some overhead here between the java `-Xmx` and the memory request.  If Picard really did try to use the full Java quota, it would exceed this amount.  (16 GiB > 16000 MiB!)

chrisamiller:
New commit addresses the points made above. Thanks for the feedback!

PR #918, opened 2020-08-04
adding germline_wgs_gvcf pipeline

apaul7:
very similar to #917 but for wgs. Will allow users to get aligned crams and GVCF files which can be used for downstream joint genotyping analysis.


apaul7:
I think I figured out rebasing for this PR but messed up #917

PR #925, opened 2020-08-06
Add padding to target regions in somatic variant calling workflows

### LINKED ISSUE #915: handle ROI wingspans in variant calling workflows
In the somatic, tumor-only, and germline workflows, interval lists are input that describe the locations being captured.  Currently, variant calls are strictly restricted to being within these target regions, which results in a fair number of false negatives (splice sites, for example).

The solution we favor is to take a input interval list that defines the target region, but then also pass a  `variant-calling-wingspan` parameter that expands those regions by the specified amount, and uses that expanded set of regions for variant calling. (using Picard Intervaltools).  

Importantly, the QC steps should still probably use the original target regions, not the expanded region

chrisamiller:
Closes #915

Removes requirement of a second interval list from somatic workflows, and instead bases the roi for variant calling on the target region plus some padding (configurable option)

### DIFF
--- a/definitions/pipelines/somatic_exome.cwl
+++ b/definitions/pipelines/somatic_exome.cwl
@@ -112,18 +112,17 @@ inputs:
           bait_intervals is an interval_list corresponding to the baits used in sequencing reagent.
           These are essentially coordinates for regions you were able to design probes for in the reagent.
           Typically the reagent provider has this information available in bed format and it can be
-          converted to an interval_list with Picards BedToIntervalList. Astrazeneca also maintains a repo
+          converted to an interval_list with Picard BedToIntervalList. Astrazeneca also maintains a repo
           of baits for common sequencing reagents available at https://github.com/AstraZeneca-NGS/reference_data
     target_intervals:
         type: File
         label: "target_intervals: interval_list file of targets used in the sequencing experiment"
         doc: |
-          target_intervals is an interval_list corresponding to the targets for the sequencing reagent.
-          These are essentially coordinates for regions designed probes for in the reagent.
-          Bed files with this information can be converted to interval_lists with Picards BedToIntervalList.
+          target_intervals is an interval_list corresponding to the targets for the capture reagent.
+          Bed files with this information can be converted to interval_lists with Picard BedToIntervalList.
           In general for a WES exome reagent bait_intervals and target_intervals are the same.
     target_interval_padding:
-        type: int?
+        type: int
         label: "target_interval_padding: number of bp flanking each target region in which to allow variant calls"
         doc: |
             The effective coverage of capture products generally extends out beyond the actual regions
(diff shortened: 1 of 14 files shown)


definitions/pipelines/somatic_exome.cwl
```diff
@@ -505,13 +515,22 @@ steps:
...
             [somalier_pairs, somalier_samples]
+
+    pad_target_intervals:
+        run: ../tools/interval_list_expand.cwl
+        in: 
+            interval_list: target_intervals
+            roi_padding: target_interval_padding
+        out:
+            [expanded_interval_list]
+
```
tmooney, line 526: We haven't been setting steps apart with whitespace on either side like this.  I'd prefer to keep the files consistent.

definitions/pipelines/somatic_exome.cwl
```diff
@@ -119,13 +119,25 @@ inputs:
         label: "target_intervals: interval_list file of targets used in the sequencing experiment"
         doc: |
           target_intervals is an interval_list corresponding to the targets for the sequencing reagent.
-          These are essentially coordinates for regions you wanted to design probes for in the reagent.
+          These are essentially coordinates for regions designed probes for in the reagent.
           Bed files with this information can be converted to interval_lists with Picards BedToIntervalList.
           In general for a WES exome reagent bait_intervals and target_intervals are the same.
+    target_interval_padding:
+        type: int?
```
tmooney, line 126: I would not mark this as optional, since the tool doing the padding requires a value.  Even when required, the default will be used if a value is not supplied.

definitions/pipelines/somatic_exome.cwl
chrisamiller: I definitely prefer some whitespace for readability, but I'll drop it for consistency's sake. (and a strong desire not to go through the repo reformatting everything!

chrisamiller:
Ready for re-review here. A test run using a new PP succeeded and showed splice site variants appearing as expected.

definitions/pipelines/tumor_only_wgs.cwl
```diff
@@ -34,8 +34,8 @@ inputs:
         type: File
     picard_metric_accumulation_level:
         type: string
-    variant_detection_intervals:
-        type: File
+    roi_intervals:
+       type: File
```
tmooney, line 38: whitespace!
```suggestion
       type: File
```

definitions/pipelines/somatic_exome_cle_gathered.cwl
```diff
@@ -38,6 +38,20 @@ inputs:
         type: File
     target_intervals:
         type: File
+        label: "target_intervals: interval_list file of targets used in the sequencing experiment"
+        doc: |
+          target_intervals is an interval_list corresponding to the targets for the sequencing reagent.
+          These are essentially coordinates for regions designed probes for in the reagent.
```
tmooney, line 44: Looks like this sentence was wholly removed from many of these inputs, but it's still here (and in three other places).

PR #934, opened 2020-08-20
Hackathon GATK4 Modifications

Sam16711:
GATK4 upgrade. We decided to keep GATK3.6 unchanged for all CombineVariants related tool CWLs since GATK4 no longer has CombineVariants command. Same case for ReadBackedPhasing command. Also keep /tools/set_filter_status.cwl unchanged because GATK4 VariantFiltration doesn’t output “##contig=<ID=chrXX” header,

Refer to https://github.com/genome/analysis-workflows/issues/24.


definitions/pipelines/germline_exome.cwl
```diff
@@ -134,9 +134,6 @@ outputs:
     verify_bam_id_depth:
         type: File
         outputSource: alignment_and_qc/verify_bam_id_depth
-    gvcf:
-        type: File[]
-        outputSource: detect_variants/gvcf
```
chrisamiller, line 139: If the gvcf stuff is being removed, are inputs like `gvcf_gq_bands: gvcf_gq_bands` still being used?

dufeiyu:
I believe @Sam16711 has a successful germline build and I am working on a HCC1395 somatic exome (cle version) testing.

definitions/pipelines/germline_exome.cwl
Sam16711: They would still be used for other pipelines using the gatk_haplotypecaller.cwl in 'GVCF' mode such as whole genome sequencing. For the single sample germline pipeline, the input for gvcfs_gq_bands is still there, but I don't believe it is actually influencing the calls as haplotypecaller is not being run in GVCF mode. I left the input so that other pipelines can still use the same haplotypecaller.cwl file.

chrisamiller:
Feiyu says: 
"Compared to current CLE version of somatic_exome CWL,  GATK4 update version does not have significant impact on recall (sensitivity) and precision (PPV). It does detect one more true-positive indel and reduces the false-positive count of SNVs (better precision). Also GATK4 update affects Hs metrics a bit, which slightly improves Hs metrics except PCT_EXC_OFF_TARGET."

chrisamiller (dismissed):
+1

definitions/tools/picard_merge_vcfs.cwl
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "Picard MergeVcfs"
+baseCommand: ["/usr/bin/java", "-jar", "/opt/picard/picard.jar", "MergeVcfs"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 40000
```
tmooney, line 9: If we're reserving this much memory for this job, should we be setting `-Xmx36g` or the like for the java options, too?

tmooney (dismissed):
Mostly looks reasonable to me!

definitions/tools/picard_merge_vcfs.cwl
jasonwalker80: Is that much RAM really required, or was that a typo and we really only need 4GB. I can't imagine needing 40GB.

Sam16711: The memory requirement was set when we were merging the chromosomal.g.vcf's that had variant calls at every position in the genome. With an exome interval list, the memory requirement should be a lot less. However, I have not tested what the low end should be. Is there a way to estimate this value?

definitions/tools/picard_merge_vcfs.cwl
```diff
@@ -0,0 +1,30 @@
...
+    - class: InlineJavascriptRequirement
+    - class: DockerRequirement
+      dockerPull: "mgibio/picard-cwl:2.18.1"
+arguments:
+    ["O=merged.vcf.gz"]
```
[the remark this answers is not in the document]
Sam16711: @tmooney Do you have a suggestion for the name?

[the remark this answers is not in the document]
tmooney: Something like how this tool does it is what I had in mind:
https://github.com/genome/analysis-workflows/blob/9daeb6b8856498d5b276a996d4be1f9582216f2e/definitions/tools/select_variants.cwl#L40-L42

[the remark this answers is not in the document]
Sam16711: Oh I see, instead of having the arguments: block. This would allow us to control the output name with a parameter then?

Thanks! I just wanted to know for reference and improvement.

dufeiyu (approved):
+1

chrisamiller (approved):
new changes seem fine (mem and file naming).  Readding my +1

definitions/tools/picard_merge_vcfs.cwl
malachig: Yes, it turns out that much memory really is needed by this tool sadly.  Current tests with real data are failing after only processing chr1 and part of chr2

johnegarza: We've run across another failure due to running out of memory; bumping it back up with #956

PR #946, opened 2020-09-01
Exposes several parameters from the somatic LLR filter

chrisamiller:
This enables tuning the pipeline to be appropriate for low-purity tumors

The somatic pipeline test runs successfully, with or without these params specified.


definitions/pipelines/somatic_exome.cwl
```diff
@@ -562,6 +577,7 @@ steps:
             normal_sample_name: normal_sample_name
             vep_custom_annotations: vep_custom_annotations
             known_variants: known_variants
+
```
tmooney, line 580: I wouldn't separate the `out` from the rest of the step definition unless we're going to pursue some large-scale formatting changes.

tmooney (dismissed):
+1 🦙

chrisamiller:
Update with whitespace fix

tmooney (approved):
+1 🧇

PR #952, opened 2020-09-16
adding docm documentation

chrisamiller:


tmooney (approved):
My understanding of "if they have even low-level evidence of a mutation" is that it establishes there exists a low but present threshold for inclusion in the VCF--i.e. if there is no evidence of a mutation they'll be excluded.  Is that correct?

I ask in order to contrast with "even if they have low-level evidence of a mutation" which would suggest their inclusion occurs even when evidence is lacking.  If you'd forced me to describe the pipeline, I would have thought this is what happens--all DOCM variants are included regardless of evidence (but are marked as DOCM_ONLY).

chrisamiller:
> There exists a low but present threshold for inclusion in the VCF--i.e. if there is no evidence of a mutation they'll be excluded.

Yes, this is the correct interpretation, and what actually happens in the docm subworkflow (see https://github.com/genome/analysis-workflows/blob/master/definitions/tools/filter_vcf_docm.cwl)

tmooney:
I don't think I'd paid much mind to this line before:
https://github.com/genome/analysis-workflows/blob/c7e906c6b266c45c030be8d5eb7784e6797cd4a9/definitions/tools/filter_vcf_docm.cwl#L80

Also, to be thorough, this one:
https://github.com/genome/docker-perl_helper-cwl/blob/e9d1862a0018f137a5adb0872b5290d6672a709f/single_sample_docm_filter.pl#L29
though this one doesn't make a mention of `DOCM_ONLY` (which would match with not having the boolean option around).

Issue #732, opened 2019-07-11
Alternative approach to making bigwig.


jasonwalker80:
We should take an alternative approach to making the bigwig file of coverage in the WGS workflow. Instead of deepTools(which seems to hang on a small subset of samples with ample RAM), we should try using BEDTools and converting to bigwig. Example docs:
https://bedtools.readthedocs.io/en/latest/content/tools/genomecov.html
https://www.encodeproject.org/software/bedgraphtobigwig/

chrisamiller:
mosdepth is going to have this natively in the next release: 
https://twitter.com/brent_p/status/1162889872455041024

chrisamiller:
Ran across this today which may be of interest here:
https://github.com/cancerit/cgpBigWig

chrisamiller:
@saimukund20 is working on this

saimukund20:
should be done by this week

chrisamiller:
#954 closed this for chipseq, which is great.  Sai, can you please add this to the two rnaseq pipelines as well?

PR #954, opened 2020-09-29
New Bigwig cwl tool - cgpbigwig

saimukund20:


definitions/tools/bam_to_bigwig.cwl
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: 'cgpBigWig Converting BAM to BigWig'
+baseCommand: ["bam2bw", "-F", "1024", "-o", "MarkedSorted.bam.bw"]
```
tmooney, line 6: It'd be nice to have the output filename derived from the input filename instead of hardcoded.  (This probably requires moving the command-line arguments to the `arguments:` section instead of the `baseCommand:` section.)

definitions/tools/bam_to_bigwig.cwl
```diff
@@ -0,0 +1,33 @@
...
+        secondaryFiles: [^.bai]
+    reference:
+        type:
+            - string
+            - File
+        inputBinding:
+            position: 2
+            prefix: '-r'
+        secondaryFiles: [.fai, ^.dict]
+
```
chrisamiller, line 28: While you're messing around with it, could you expose another option or two?
https://github.com/cancerit/cgpBigWig/blob/develop/cwls/bam2bw.cwl exposes some additional options that seem like they'd be worthwhile.  Specifically, the option to handle overlapping reads seems important (and should probably be enabled by default, I think). 

Also, what's up with the scaling factor input to this tool?
```
-S  --scale-log10 [float]                         A scale factor to multiply to output [default: -1912244352]
```
Why is that the default value, and is that appropriate for both human and mouse here?

chrisamiller:
linking to issue: 
https://github.com/genome/analysis-workflows/issues/732

definitions/tools/bam_to_bigwig.cwl
chrisamiller: we've verified that despite confusing docs, the scaling factor is not an issue and shouldn't be exposed here

chrisamiller (approved):
+1

tmooney (approved):
+1 While it'd be my personal preference to avoid using capital letters in step names, this looks like it'll work!

PR #956, opened 2020-10-01
Updating ramMin and baseCommand of picard_merge_vcfs.cwl to provide m…

Matthew-Mosior:
…ore memory to java.

### DIFF
--- a/definitions/tools/picard_merge_vcfs.cwl
+++ b/definitions/tools/picard_merge_vcfs.cwl
@@ -3,10 +3,10 @@
 cwlVersion: v1.0
 class: CommandLineTool
 label: "Picard MergeVcfs"
-baseCommand: ["/usr/bin/java", "-jar", "/opt/picard/picard.jar", "MergeVcfs"]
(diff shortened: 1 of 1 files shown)


johnegarza (approved):

tmooney (approved):
+1 seems like it'll work!  Do we know it needs so much memory or was the previous issue from having the memory request be lower than the Java heap limit?

Matthew-Mosior:
That's a good question.  I know our current immuno.cwl have succeeded past this point, but not sure if it was increasing ramMin, baseCommand with the java heap limit increase, or both.

definitions/tools/picard_merge_vcfs.cwl
```diff
@@ -3,10 +3,10 @@
 cwlVersion: v1.0
 class: CommandLineTool
 label: "Picard MergeVcfs"
-baseCommand: ["/usr/bin/java", "-jar", "/opt/picard/picard.jar", "MergeVcfs"]
+baseCommand: ["/usr/bin/java", "-jar", "-Xmx38g", "/opt/picard/picard.jar", "MergeVcfs"]
```
tmooney, line 6: I realized there's a bug here--the memory param got stuck between `-jar` and its value.  There is a fix for that as part of #960.

PR #963, opened 2020-10-09
Updating versions of several tools

chrisamiller:
Tested by running the somatic exome test.  The sole difference was one additional dinucleotide variant called from mutect and passed through.  This is not unexpected, given some of the updates that have been made to Mutect in that area.

PR #964, opened 2020-10-16
Updating tool path to work in new docker image

johnegarza:
#963 switched Picard docker images from an mgibio maintained image to the official Broad image. The mgibio image included the Picard jar in two different locations, but it's only present in one in the Broad image. This patch updates the path to the one present in the Broad image; all other Picard cwls use the path present in both images.


chrisamiller (approved):
+1

Issue #943, opened 2020-09-01
implement blocklist_filter to sv_detect_variant for Germline WGS


jasonwalker80:
Feiyu - 
> I have already had the filter_vcf_blocklist_bedpe.cwl working as tool cwl and my version of germline_wgs_cle_analysis 
> pipeline cwl. But I want to tweak it into sv_detect_variant as optional so both cle and non-cle user can use

dufeiyu:
The PR is merged. Resolve this issue.

PR #965, opened 2020-10-19
Add SV blocklist vcf filter

dufeiyu:
An extra step to filter SVs listed in blocklist bedpe is needed for germline_wgs.cwl.  Refer to https://jira.ris.wustl.edu/browse/CI-487 and https://github.com/genome/analysis-workflows/issues/943

### DIFF
--- a/definitions/tools/filter_sv_vcf_blocklist_bedpe.cwl
+++ b/definitions/tools/filter_sv_vcf_blocklist_bedpe.cwl
@@ -17,19 +17,12 @@ requirements:
           set -eou pipefail
           set -o errexit
           
-          INPUT_VCF="$1"
-          BL_BEDPE="$2"
-          SLOPE="$3"
-          OUT_BASE="$4"
-          
-          #BASE=`basename $1`
-          #NAMEROOT=`echo $BASE | perl -pe 's/.vcf(.gz)?$//g'`
-          
-          if [[ "$BL_BEDPE" == 'NONE' ]]; then
-              /usr/local/bin/bedtools sort -header -i "$INPUT_VCF" > $OUT_BASE.vcf 
-              /opt/htslib/bin/bgzip $OUT_BASE.vcf
-              /usr/bin/tabix -p vcf $OUT_BASE.vcf.gz
-          else
+          if [[ "$#" == 4 ]];then # blocklist_bedpe is passed.
+              INPUT_VCF="$1"
+              BL_BEDPE="$2"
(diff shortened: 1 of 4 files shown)


definitions/tools/filter_sv_vcf_blocklist_bedpe.cwl
```diff
@@ -0,0 +1,76 @@
...
+          fi
+
+inputs:
+    input_vcf:
+        type: File
+        inputBinding:
+            position: 1
+        doc: "vcf file to filter"
+    blocklist_bedpe:
+        type: string
```
tmooney, line 55: Would this make more sense as a `File` input?  Additionally, it seems like it would make more sense to make it optional rather than expect the string `NONE`.  If there's a reason it can't use the more common way of omitting an input, it'd be nice to have an explanation in the `doc` line about the unique requirement.  (One way to have optional parameters might be to use `getopts` like here: https://github.com/genome/analysis-workflows/blob/c2276ac6374436334e5f0ce3a0197a751c22b1bc/definitions/tools/sequence_align_and_tag.cwl#L29)

dufeiyu: I agree. I couldn't find a better way to handle optional parameter **without** default for embedded shell and therefore took this compromised approach. Besides `getops`, what do you think this way ?  
https://github.com/genome/analysis-workflows/blob/ece70ac30cd87100a70f7dc64d08fa72724e9416/definitions/tools/docm_gatk_haplotype_caller.cwl#L20
If the optional parameter input is not passed, the positions of inputs behind will be reduced by 1.

tmooney: Sure! Parameter counting works, too.  I suggested `getopts` because it looked like a few other optional arguments already in the tool.  Once multiple arguments could be omitted it's harder to track them solely by counting.  (If they are required but have defaults, it should be safe to remove the `?` from their definitions in the tool CWL.)

dufeiyu:
@tmooney  I change the input type of blocklist_bedpe to File optional as what we discussed.

tmooney (approved):
+1 I'm trusting that the bash script does what you want it to, but it generally looks good to me!

dufeiyu:
@apaul7 Can you review this ? Thanks.

apaul7 (approved):
+1
Once Cromwell supports CWL v1.2 we can start using conditional steps!

PR #971, opened 2020-11-20
Adding bigwig creation to the Chipseq / ATAC seq human pipeline

saimukund20:


definitions/pipelines/immuno.cwl
```diff
@@ -12,6 +12,11 @@ requirements:
     - class: SubworkflowFeatureRequirement
 inputs:
     #rnaseq inputs
+    reference_rnaseq:
```
tmooney, line 15: Is it necessary for this to be a separate reference than the existing `reference` input?

definitions/pipelines/rnaseq.cwl
```diff
@@ -155,3 +163,10 @@ steps:
             bam: index_bam/indexed_bam
         out:
             [metrics, chart]
+    cgpBigWig_bamcoverage:
```
tmooney, line 166: It'd be my personal preference if we keep step names to snake_case rather than capitalizing some letters.

definitions/pipelines/immuno.cwl
saimukund20: I dont see any reference (.fa) passed to the RNA seq pipeline. So to make sure the RNA seq pipelines has it i had to add it to the immuno pipeline as well.

definitions/pipelines/rnaseq.cwl
saimukund20: Ive made this change.

definitions/pipelines/immuno.cwl
chrisamiller: hrmm - you're right.  The "reference_index" param right below that is a pointer to the hisat indexes (like `/gscmnt/gc2560/core/GRC-human-build38_human_95_38_U2AF1_fix/rna_seq_annotation/hisat2.1.0_index/GRCh38DH`, which don't necessarily have the fasta adjacent to them.  If we swap this out for the star pipeline in the future, it'll get passed the `star_genome_dir` (which contains the fasta IIRC) and we could drop this param, but it does seem necessary here, I think.

tmooney: To clarify, I am not asking about whether a reference needs to be passed to rnaseq.  I'm wondering if this parameter could be used:
https://github.com/genome/analysis-workflows/blob/b465f0da2806ddb6df481409541d13288ccb40ec/definitions/pipelines/immuno.cwl#L57-L61
rather than creating a new `reference_rnaseq` input.  The existing `reference` input is used for several of the other subworkflows in `immuno.cwl`.

saimukund20: Right that makes sense. Fixed it.

definitions/pipelines/rnaseq.cwl
```diff
@@ -155,3 +163,10 @@ steps:
             bam: index_bam/indexed_bam
         out:
             [metrics, chart]
+    cgpbigwig_bamcoverage:
```
chrisamiller, line 166: In some coverage tools (bedtools for example), you need to specify special parameters to handle split reads, so that a read split across two exons doesn't get counted through the entire intron. Is that the case with this tool?  (Is that what the `-a  --overlap   Use overlapping read check` param is for?)  Can you grab a small test case, and confirm whether this is handling such a case properly?

saimukund20: Yes you are right, the -a  --overlap checks for those overlapping reads. Should I add that?

chrisamiller: For the RNA-Seq calculations, yes, please!

Issue #928, opened 2020-08-06
Add target regions for consistency in tumor_only_wgs.cwl


chrisamiller:
Right now `tumor_only_wgs.cwl` doesn't take in interval lists in the same way that the tumor-only exome, or somatic pipelines do.  Instead of target/bait intervals, it takes in `qc_intervals` and `interval_list`. It should probably be modified to be consistent with the others.

PR #974, opened 2020-12-09
updates intervals, adds missing params and test yaml

chrisamiller:
Updates interval list params to more closely resemble those in the other workflows.  Also exposes a few missing params, and adds a test yaml

### DIFF
--- a/example_data/tumor_only_wgs.yaml
+++ b/example_data/tumor_only_wgs.yaml
@@ -1,0 +1,102 @@
+---
+reference:
+  class: File
+  path: somatic_inputs/hla_and_brca_genes.fa
+
+sequence:
+  - sequence:
+      bam:
+        class: File
+        path: unaligned_subset_bams/tumor/2895499223.bam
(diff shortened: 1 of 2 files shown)


chrisamiller:
closes https://github.com/genome/analysis-workflows/issues/928

example_data/tumor_only_wgs.yaml
```diff
@@ -0,0 +1,102 @@
...
+- set
+variants_to_table_genotype_fields:
+- GT
+- AD
+- AF
+- DP
+
+
+
+
```
tmooney, line 102: There's some inconsistent whitespace in this file.  Some items are separated by one blank line, some by two, some by none, and four appear here at the end.

tmooney (commented):
Was this PR meant to also include some changes to the CWL itself?  Some of the params in this example YAML don't match what's in `tumor_only_wgs.cwl`.

chrisamiller:
ugh - thank you, the cwl didn't get pushed.  Updated the whitespace as well

gschang:
+1

tmooney (approved):
+1, if I were to say anything I might say adding doc lines to some of the inputs makes me wonder if we should add some to the other new inputs being added here!

PR #975, opened 2020-12-09
Remove stray eXecute bit from a few CWL files.

tmooney:
Three CWL files in our repo had the execute bit set on them.  While this is allowed by the CWL spec, I don't think there was any intention behind setting it for specifically these three.  Absent a reason to do otherwise, I'd prefer that we keep them all consistent.


chrisamiller (approved):
+1

apaul7 (approved):
+1

PR #976, opened 2020-12-09
updating kallisto version

chrisamiller:
produces no changes to gene/transcript abundance on our test data, and the only difference in the fusion outputs is some minor changes to ordering of lines in the files.  Should be a seamless upgrade


apaul7 (approved):
+1

PR #977, opened 2020-12-12
Bam to cram support

smk5g5:
Bam to CRAM support added to rnaseq star fusion cwl


definitions/pipelines/rnaseq_star_fusion.cwl
```diff
@@ -192,3 +192,16 @@ steps:
             bam: mark_dup/sorted_bam
         out:
             [metrics, chart]
+    bam_to_cram:
+        run: ../tools/bam_to_cram.cwl
+        in:
+          reference: cdna_fasta
```
chrisamiller, line 198: The cdna fasta is not the same as the reference sequence fasta.  It contains one fasta entry for each transcript.  The reference fasta is needed here, either derived from the star directory, or passed in explicitly

smk5g5: Mybad!

tmooney (approved):
+1 🦀

chrisamiller (approved):
+1  looks good!

Issue #948, opened 2020-09-02
Review software for upgrades


chrisamiller:
Prior to pushing out a 2.0 release and re-benchmarking, we should review key software components and bump versions to the latest and greatest when it makes sense. These don't necessarily _need_ to be updated, but if we are going to do so, before a release is the right time.   

A preliminary list of software to check, based on docker images:

- [ ] smoove
- [ ] somalier
- [ ] bedtools
- [ ] cnvkit
- [ ] vatools
- [ ] kallisto
- [ ] star
- [ ] starfusion
- [ ] biscuit
- [ ] manta
- [ ] strelka
- [ ] picard
- [ ] pizzly
- [ ] sambamba
- [ ] samtools
- [ ] vep
- [ ] vt
- [ ] cellranger
- [ ] hisat
- [ ] stringtie
- [ ] bcftools
- [ ] fgbio
- [ ] mutect

Feel free to add things I've missed below, or to check these off when you've confirmed that a) they have been upgraded or b) they don't need to be (add comments confirming below)

chrisamiller:
Notes on a few of these.  Switching to "official" docker containers or those in bioconda/biocontainers seems like a good plan, especially when it reduces dependency on our own images or dockerhub.

#kallisto: 
quay.io/biocontainers/kallisto:0.46.1--h4f7b962_0
#should be no index change - test

#hisat
quay.io/biocontainers/hisat2:2.2.1--he1b5a44_2

#star-fusion
quay.io/biocontainers/star-fusion:1.9.0--1

#stringtie:
quay.io/biocontainers/stringtie:2.1.2--h7e0af3c_1

#biscuit:
#may have some param changes (?)
#new URL https://github.com/huishenlab/biscuit
quay.io/biocontainers/biscuit:0.3.16.20200420--h2b0c03c_2 


#somalier
#Looks like minor bugfixes. A version bump and a quick sanity check on a pair of bams oughta be fine
#hosted on dockerhub - brentp/somalier,  seems like we should create our own to remove dependency on other people's dockerhub, at least

mutect: 
seems like a bump to newest version `4.1.8.1` would be easy/fine - couple of bugfixes

picard:
# seems like these are all bugfixes, just a bump is probably fine
quay.io/biocontainers/picard:2.23.3--0

chrisamiller:
[...]

PR #978, opened 2020-12-14
Upgrade fgbio to 1.3.0

tmooney:
This switches away from our fat DNA alignment image to an fgbio image on biocontainers as well.

This is part of #948.


chrisamiller (approved):
I confirm that the ﻿docker image exists, the path is correct for the commands, and that the clipping mode param matches up with the new setting (it's already Hard by default, but I'm all for being explicit!)

PR #980, opened 2020-12-15
Fix filter_vcf_cle.cwl for CLE VCF filtration in the immuno CWL pipeline

gschang:
The filter_vcf_cle step ("tools/filter_vcf_cle.cwl") in "subworkflows/filter_vcf.cwl" was implemented to filter out any somatic variants that haven't been validated by the CLE somatic pipeline. This step is required for some clinical cases when we make a pVACseq analysis using the immuno CWL pipeline.

However, the current filter_vcf_cle.cwl doesn't work, ending up with writing the same output VCF as the input VCF. This results in feeding nonCLE variants to the pvacseq analysis workflow in CLE-required clinical cases.  In fact, the script "/usr/bin/docm_and_coding_indel_selection.pl" (dockerPull: "mgibio/cle:v1.3.1") that the tools/filter_vcf_cle.cwl uses isn't the right one.

In this PR,

(1) I wrote a new script to replace "/usr/bin/docm_and_coding_indel_selection.pl" by adding a new entry in InitialWorkDirRequirement class. The new script formats the final VCF as an input for pVACseq tools.

In detail, the goal is to feed only CLE-validated somatic variants to pVACseq analysis for CLE-required clinical cases. It updates the FILTER status of the variant whose FILTER status is PASS but that variant is not validated by the CLE somatic pipeline. There are many ways to achieve this goal, and this approach looks best in the current workflow context. The filter_vcf_cle step (tools/filter_vcf_cle.cwl) is implemented in "pipelines/somatic_exome.cwl" pipeline.

(2) I update the inputBinding definition of the "filter" input. This now enables us to use this filter_vcf_cle as an option. Actually, this "filter" input does nothing in the current "tools/filter_vcf_cle.cwl".

(3) minor update with the latest docker, "mgibio/cle:v1.4.2"

### DIFF
--- a/definitions/tools/filter_vcf_cle.cwl
+++ b/definitions/tools/filter_vcf_cle.cwl
@@ -2,22 +2,158 @@
 
 cwlVersion: v1.0
 class: CommandLineTool
-label: "cle_annotated_vcf_filter"
-baseCommand: ["/usr/bin/perl", "/usr/bin/docm_and_coding_indel_selection.pl"]
+label: "Filter variants from the VCF annotated with CLE-validated somatic VCF"
+baseCommand: ["/usr/bin/perl", "cle_filter.pl"]
 requirements:
     - class: DockerRequirement
-      dockerPull: "mgibio/cle:v1.3.1"
+      dockerPull: "mgibio/cle:v1.4.2"
     - class: ResourceRequirement
       ramMin: 4000
-arguments: [$(inputs.vcf.path), $(runtime.outdir)]
+    - class: StepInputExpressionRequirement
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: 'cle_filter.pl'
+        entry: |
+            #!/usr/bin/perl
+            
+            # This script intends to prepare a somatic VCF input file 
+            # for pVACseq analysis with clinical trials.
+            # For example, it updates the FILTER status of the variant whose FILTER status is PASS
+            # but, if that variant is not validated by the CLE somatic pipeline.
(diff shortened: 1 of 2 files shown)


chrisamiller (approved):
+1 seems to do what it says it should do

gschang:
Thank you for this comment, Chris. Jasreet arranged a meeting to discuss with John tomorrow and I will update you more.

gschang:
John, Jasreet, and I discussed this issue this afternoon. We're brainstorming to move forward for the best practice in the immuno workflow.

gschang:
Moreover, I added a minor upgrade for "tools/pvacseq.cwl" in this PR. "ramMin: 16000" looks like minimum, and the pvacseq job can be choked as the memory runs out. We usually recommend 32000 for this parameter, which I am using for our clinical studies.

dufeiyu:
Hi Gue Su,  filter_vcf_cle.cwl that runs docm_and_coding_indel_selection.pl was specifically made for CLE purpose (selecting docm variants and coding indel) by Dave's request a long while ago.  If you want extra filter for pVACSeq purpose, you probably need make a new filter_vcf_xxx.cwl and add to subworkflow/filter_vcf.cwl as optional if needed instead of rewriting filter_vcf_cle.cwl.

gschang:
Hi @dufeiyu. Thank you for your feedback. As John, Jasreet, and I discussed today, I am going to close this PR without merging, in order to keep this "filter_vcf_cle" step as it is. So you and Dave can work more to implement a CLE-customized filtration later.

PR #981, opened 2020-12-16
Added bam/fastq input compatability for alignment_exome_nonhuman.cwl

Sam16711:
Added bam/fastq input compatibility for alignment_exome_nonhuman.cwl pipeline. This was done using sequence_to_bqsr_nonhuman.cwl. The is name is misleading as bqsr is not actually ran for this subworkflow.

### DIFF
--- a/definitions/pipelines/alignment_exome_nonhuman.cwl
+++ b/definitions/pipelines/alignment_exome_nonhuman.cwl
@@ -7,17 +7,21 @@ requirements:
     - class: SchemaDefRequirement
       types:
           - $import: ../types/labelled_file.yml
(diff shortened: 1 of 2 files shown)


tmooney (commented):
This change also requires updating a few locations that use this pipeline, e.g.:
https://github.com/genome/analysis-workflows/blob/3a287b7cb6162cdea79865235d224fea45963d87/definitions/pipelines/somatic_exome_nonhuman.cwl#L263

tmooney (approved):
+1

I opened #987 about some of the documentation for sequence data.  Since this PR copied the docs from existing inputs, we can merge this and then address all of them together.

PR #982, opened 2020-12-17
Convert UMI alignment pipelines to use `sequence_data` custom type.

### LINKED ISSUE #969: Write equivalent sequence_align_and_tag.cwl for umi duplex and umi molecular pipelines.
Issue in reference to #935 . In order to support bam/fastq input for the umi duplex and umi molecular pipelines, an equivalent to sequence_align_and_tag.cwl will have to be written. Currently, both pipelines require fastq files that are eventually converted to bams using `../tools/fastq_to_bam.cwl`.

tmooney:
The UMI pipelines depend on having BAMs for input, but previously our outer pipeline CWLs required FASTQ inputs which were then immediately converted to BAM.  Now if a BAM is available it can be used as an input to avoid this conversion step.

Closes #969 in spirit.

### DIFF
--- a/definitions/pipelines/alignment_umi_duplex.cwl
+++ b/definitions/pipelines/alignment_umi_duplex.cwl
@@ -4,21 +4,16 @@ cwlVersion: v1.0
 class: Workflow
 label: "umi duplex alignment fastq workflow"
 requirements:
+    - class: SchemaDefRequirement
+      types:
+          - $import: ../types/sequence_data.yml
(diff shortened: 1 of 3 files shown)


jasonwalker80 (approved):
+1

tmooney:
I just merged this, but it occurs to me it's possible that someone using this tool was hoping the RG mentioned in the sequence would get propagated to the unaligned BAM file.  Supporting that case would be more complicated than the `cp` used here.

PR #986, opened 2020-12-30
Updated version of samtools to 1.11

smk5g5:
updated samtools version to 1.11 (uses quay.io image)


chrisamiller:
I assume that the quay.io container contains bgzip, which is needed by a couple of these steps?  Nothing here that you're altering seems to use tabix, which is the other thing in the container that's being replaced.

smk5g5:
> I assume that the quay.io container contains bgzip, which is needed by a couple of these steps? Nothing here that you're altering seems to use tabix, which is the other thing in the container that's being replaced.

Yes it does contain bgzip I have changed the path to /usr/local/bin/bgzip in the bgzip cwl. I think there are a couple of shell scripts that I may need to change as well that I haven't touched!

chrisamiller (dismissed):
+1

tmooney (changes requested):
As you noted, some of the shell scripts need updating before this can be used.  Since this got an approval, marking request changes to avoid accidental merges :smile:

chrisamiller:
ugh - I thought I checked them all, but `definitions/tools/vcf_sanitize.cwl` needs an update, since it requires tabix. Most of the others are fine, I think...

tmooney:
Noting that `index_vcf.cwl` needs its tabix path updated as well.

tmooney (approved):
+1 🗝️

Issue #979, opened 2020-12-15
Fix filter_vcf_cle.cwl for CLE VCF filtration in the immuno CWL pipeline


gschang:
I am working for testing the latest immuno workflow, by discussing with John and Jasreet.

The filter_vcf_cle step (tools/filter_vcf_cle.cwl) in subworkflows/filter_vcf.cwl was implemented to filter out any somatic variants that haven't been validated by the CLE somatic pipeline. For example, this step is required for some clinical cases when we make a pVACseq analysis using the immuno CWL pipeline. However, the current filter_vcf_cle.cwl doesn't work, ending up with writing the same output VCF as the input VCF.

Below, I have created a new PR to suggesting fixing this issue.

Fix filter_vcf_cle.cwl for CLE VCF filtration in the immuno CWL pipeline
#980

gschang:
I closed the previous PR (#980) without merging, as Feiyu made that "filter_vcf_cle" step for CLE. John, Jasreet, and I discussed a plan to fix this issue, and we are going to create a new step in the immuno workflow as Feiyu suggests. I will update this.

gschang:
This issue is now moved to a new PR, "Limit pvacseq input to passing cle and pipeline variants" #988.

PR #988, opened 2021-01-06
Limit pvacseq input to passing cle and pipeline variants

johnegarza:
This PR is a fix for #979. 
First, it updates the annotation step of the workflow: previously, any variant present in the cle vcf (if provided) matching a variant in the pipeline vcf was annotated in the pipeline vcf; now, only passing variants in the cle vcf are considered. The annotation label was updated from `PREVIOUSLY_DISCOVERED` to `CLE_VALIDATED` to more clearly indicate the source of the annotation, and the name of the input file variable within the workflows was updated accordingly. The behavior in the event that no cle vcf is provided remains unchanged: the input vcf is simply copied to be passed along to the next step unchanged.
Next, it adds an intersection step just before pvacseq is run. If a cle vcf is provided, this step takes the intersection of passing cle variants and the pipeline vcf. If no cle vcf is provided, as with the above step, the input pipeline vcf is copied. No filtering is necessary because pvacseq is run with the `--pass-only` flag, so only passing variants in the input vcf are considered. This step was not included in the above annotation step because that step is nested several layers into the workflow and would affect other pipelines if modified, and the resulting vcf is currently only intended for use in the immuno pipeline.

### DIFF
--- a/definitions/pipelines/detect_variants.cwl
+++ b/definitions/pipelines/detect_variants.cwl
@@ -129,10 +129,10 @@ inputs:
     vep_custom_annotations:
         type: ../types/vep_custom_annotation.yml#vep_custom_annotation[]
         doc: "custom type, check types directory for input format"
-    cle_variants:
+    validated_variants:
         type: File?
         secondaryFiles: [.tbi]
-        doc: "Previously discovered variants to be flagged in this pipelines's output vcf"
+        doc: "An optional VCF with variants that will be flagged as 'VALIDATED' if found in this pipeline's main output VCF"
 outputs:
     mutect_unfiltered_vcf:
         type: File
@@ -386,7 +386,7 @@ steps:
                     }
                     return('gnomAD_AF');
                 }
-            cle_variants: cle_variants
+            validated_variants: validated_variants
         out: 
             [filtered_vcf]
     annotated_filter_bgzip:
--- a/definitions/pipelines/detect_variants.cwl
+++ b/definitions/pipelines/detect_variants.cwl
@@ -129,7 +129,7 @@ inputs:
     vep_custom_annotations:
         type: ../types/vep_custom_annotation.yml#vep_custom_annotation[]
         doc: "custom type, check types directory for input format"
-    known_variants:
+    cle_variants:
         type: File?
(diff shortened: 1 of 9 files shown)


chrisamiller:
Few quick thoughts:

1) This description could probably be made more explicit, since this is kind of confusing:
"Previously discovered variants to be flagged in this pipelines's output vcf"

2) Would calling it "VALIDATED" instead of "CLE_VALIDATED" make things more general?  I can maybe picture this being used in other non-cle contexts - only keeping vars when they're called by two centers, for example? (which would lend itself towards keeping it as "known_variants" or "validated_variants" or something as opposed to "cle_variants".  Open for discussion on whether this would make things more or less confusing!

3) Might be good to add a bit more description to the `intersect_known_variants.cwl` so it's clear what's going on there - the bcftools commands can get rather cryptic.

gschang:
This PR started from discussion on the immuno workflow, because many clinical cases require CLE validation. The current immuno.CWL feeds all somatic variants to the pvacseq subworkflow. As Chris points out, this issue turns out to need more discussing, because of two pipeline aspects--between (1) generalization for variant annotation and (2) customization for variant filtration for CLE purpose.

(Hoping for clarification, please correct me if I am wrong) What John and we propose in this PR has two parts--(1) annotation label update, currently as PREVIOUSLY_DISCOVERED, and (2) variant intersection with CLE VCF as a new workflow "intersect_known_variants.cwl". The first part impacts the variant annotation in the somatic_exome workflow, so we should consider the big picture for this pipeline as Chris mentioned. We may need more discussing and coordinating with Chris, Dave, Feiyu, and so on. The second part is intended to fix the issue that we have for CLE validation, which we plan to implement in the immuno workflow.

In sum, we can limit the scope of this PR to facilitate issue solving. I think that the first part needs more time and discussion, to figure out the big pictures, where I don't see any end point yet. Many contexts are going. In fact, I closed the PR #979 because my solution doesn't fit this overall framework, as Feiyu pointed out. To make it simple, the second part is what we need to fix this problem in the immuno workflow at this moment. However, we are always open to any feedback here.

chrisamiller (approved):
+1

Issue #987, opened 2021-01-05
"sequence" type is not a YAML file


tmooney:
Several places in our documentation refer to "sequence" as a YAML file.  For example:

https://github.com/genome/analysis-workflows/blob/3a287b7cb6162cdea79865235d224fea45963d87/definitions/pipelines/somatic_exome.cwl#L39-L47

This sounds like the sequence should be passed as a YAML file, when in fact this input requires an array of the "sequence" data structure with read group information and BAM or FASTQ files.

PR #989, opened 2021-01-12
Documentation updates (mostly to "sequence" input fields)

tmooney:
This cleans up the documentation for "sequence" inputs (#987) and then propagates the updated documentation to other pipeline CWLs that have sequence inputs.  Also fixes a few typos and redundancies in other somatic documentation.


chrisamiller (approved):
+1 better docs are always welcome!

PR #997, opened 2021-01-29
Somatic fix

chrisamiller:
The new containers added here either don't have needed tools (sanitize) or segfault (merge). Rolling these back to older versions.


gschang:
(+1) This needs fixing ASAP.

Thank you for this work, @chrisamiller. As you pointed, I had the same problem in ../tools/vcf_sanitize.cwl, but I didn't dig it up yesterday. I just know that this problem is also owing to the new docker container. As updated, I am testing somatic_exome workflow with example data. I will keep you posted if I find another issues.

tmooney (dismissed):
+1, though I might prefer if you cherry-picked the commit that matters onto a clean branch 😄

tmooney:
What issues did you run into on the sanitize step?  It looks like tabix and bgzip are both available in the newer image, albeit at different paths.

chrisamiller:
Sorry - my notes got mixed up.  tabix and bgzip do indeed exist in the quay container, but perl does not.

chrisamiller:
Please hold off on this review - I'm doing some additional testing

chrisamiller:
I have screwed up this branch by merging in changes from my other PR - starting over

PR #998, opened 2021-02-01
Fixes merging issue

chrisamiller:
Using the current somatic_exome workflow and the example data, BQSR crashes.  This error can be traced back to the fact that when merging two bams with samtools, it does not add header lines for each RG group, as expected.  Sambamba isn't a drop-in replacement because it can't handle merging unsorted data.  So in the case where we're merging unsorted aligned (as in sequence_to_bqsr.cwl and several other pipelines), then picard's MergeSamFiles is the right solution. 

Rather than make yet another merge_bams cwl, this solves the problem by switching to the right solution for several different merging contexts.


definitions/tools/merge_bams.cwl
```diff
@@ -3,26 +3,81 @@
...
+            #if there is only one bam, just copy it and index it
+            if [[ $NUM_BAMS -eq 1 ]]; then
+                cp "$BAMS" "$OUTFILENAME";
+            else
+                if [[ $SORTED == "true" ]];then
+                    /usr/bin/sambamba merge -t "$NTHREADS" "$OUTFILENAME" "$BAMS"
+                else #unsorted bams, use picard
+                    MEM=`echo "$MEM/1000" | perl -nae 'print eval $_'`
+                    cmd="java -jar -Xmx"
+                    cmd+="$MEM"
```
tmooney, line 50: I don't believe it's safe to give all of the memory to java's heap.  But even if it were, it might be better to use a unit signifier letter to avoid the math.  (Like `"-Xmx" . $MEM . "m"` if it were a perl expression).

definitions/tools/merge_bams.cwl
```diff
@@ -3,26 +3,81 @@
...
+                    cmd+="g /opt/picard/picard.jar MergeSamFiles OUTPUT=$OUTFILENAME ASSUME_SORTED=true USE_THREADING=true SORT_ORDER=unsorted VALIDATION_STRINGENCY=LENIENT"
+                    for i in $BAMS;do #this assumes no spaces in filenames, but a space in a filename is a space in one's soul
+                      cmd+=" INPUT=$i"
+                    done
+                    `$cmd`;
```
tmooney, line 55: If `$cmd` were building up a bash array of the parameters, we could handle spaces properly and not need to use backticks here.

chrisamiller:
Cleaned this up - thanks for the feedback @tmooney

gschang:
I am testing somatic_exome workflow with the example data. I cloned the current analysis-workflows Git repository as of today (#999 PR merged). However, my test run crashes in "./tools/bqsr.cwl" tool. I tracked it down and this PR sounds like related with my BQSR crash, which may be because of BQSR step upgrade with GATK v4.1.8.1.

@chrisamiller and @tmooney, though I am now digging up, I wonder if my test somatic_exome run is stuck with the same issue in this PR. If this PR is merged, will this BQSR crash issue be gone?

definitions/tools/merge_bams.cwl
```diff
@@ -3,26 +3,85 @@
...
+
+            BAMS=("${@:$OPTIND}")
+            NUM_BAMS=${#BAMS[@]}
+
+            #if there is only one bam, just copy it and index it
+            if [[ $NUM_BAMS -eq 1 ]]; then
+                cp "$BAMS" "$OUTFILENAME";
+            else
+                if [[ $SORTED == "true" ]];then
+                    /usr/bin/sambamba merge -t "$NTHREADS" "$OUTFILENAME" "$BAMS"
```
tmooney, line 49: I think `$BAMS` here needs to be updated to refer to the array.
```suggestion
                    /usr/bin/sambamba merge -t "$NTHREADS" "$OUTFILENAME" "${BAMS[@]}"
```
(It's fine in the `cp` case since it'll include the first entry, but that could also be updated to be an explicit reference to the first entry if desired.)

gschang: Thank you for this review. I am listening.

chrisamiller:
Yes, @gschang this fixes the bqsr issue

Thanks, tmooney - oversight on my part (and lack of testing of that particular path through the tree). Your suggested change was merged.

tmooney (approved):
+1, though I'm still not convinced it's safe to give 100% of the runtime memory to Picard.

gschang:
Thank you so much for this work, Tom and Chris. I have just made a Merge pull request for this PR. I am going to run another test with somatic_exome workflow examples with this change.

gschang:
As a quick update on this BQSR issue, my somatic_exome workflow test is running, and I confirmed that this PR solved this crash problem.

gschang:
Hi @tmooney and @chrisamiller. I found that the "definitions/tools/merge_bams.cwl" does not work with the current "definitions/pipelines/rnaseq.cwl". In brief, this is because the "definitions/tools/merge_bams.cwl" doesn't sort sequencing reads when merging aligned BAMs.

In detail, the current "rnaseq.cwl" uses "merge_bams.cwl" in order to merge aligned BAMs from RNA sequencing data. However, This merge_bams.cwl step results in an unsorted BAM, when I test it with the current rnaseq examples. This unsorted BAM crashes the next step "index_bam" in "rnaseq.cwl" (because it requires sorted BAMs). So the rnaseq workflow ends up broken.

The following directory is an actual example: 

/gscuser/gchang/work2/14_BGA-WALKER/210203-cwltool2examples-958/1.analysis-workflows-test/example_data/test/rnaseq.cwl/0d1578d0-c870-4280-8155-bb1065ce436a/call-merge/execution/merged.bam


I used the following RNA-seq examples, after updating with SAM header lines. (The original ones don't have any header lines.)

(1) /gscuser/gchang/work2/14_BGA-WALKER/210203-cwltool2examples-958/1.analysis-workflows-test/example_data/rnaseq/gerald_H3MYFBBXX_4_GCCAAT_1k.bam
(2) /gscuser/gchang/work2/14_BGA-WALKER/210203-cwltool2examples-958/1.analysis-workflows-test/example_data/rnaseq/gerald_H3MYFBBXX_5_GCCAAT_1k.bam


In fact, the "definitions/tools/merge_bams.cwl" works fine in "definitions/pipelines/somatic_exome.cwl" (as described in this PR), because the current somatic_exome.cwl sorts the merged BAM next. However, the current rnaseq.cwl doesn't work with merge_bams.cwl.

chrisamiller:
@gschang please make a new issue with this information, and then coordinate with @sridhar0605 (who is hacking on the RNAseq pipelines right now). The solution is probably going to be to insert a "sort_bams" step in between merging and indexing.

sridhar0605:
I cannot access those workflow paths @gschang could you please change perms. [This](https://github.com/genome/analysis-workflows/blob/master/definitions/tools/merge_bams.cwl#L47) code block should take care of sorted or unsorted bams.

gschang:
@chrisamiller. Your comment make sense to me. I will coordinate with Sridhar to fix this problem in the rnaseq workflow. I will makes a new issue.

PR #999, opened 2021-02-02
revert breaking changes to vcf_santize and samtools_merge

chrisamiller:
See https://github.com/genome/analysis-workflows/pull/997 for the full discussion before I screwed up my branch.  tldr:  perl is missing from the sanitize container and the new version of samtools merge segfaults.  Rolling both changes back.


gschang:
(+1) This needs to be done. The current workflow examples don't work and our goal is to make working workflow YAML examples.

johnegarza (approved):

PR #1000, opened 2021-02-03
updates to somatic and bisulfite test yamls

chrisamiller:
Changes the somatic and bisulfite yamls to match pipeline changes. Also adds fastq input testing


gschang:
(+1) When this PR is merged, I am going to make another test run with these changes. As we discussed at today's meeting, I was working to update somatic_exome YAML example.

I was stuck with other steps that didn't work, when I arranged an actual test run with my YAML example. I will coordinate with Chris's effort here to complete this.

tmooney (approved):
+1, finally went to the trouble of adding newlines between every entry in the somatic YAML! I'm not sure I understand the reordering that took place there at the same time, but it also seems fine 🙂

PR #1001, opened 2021-02-04
Handling picard memory in the merge step

chrisamiller:
I don't love that this requires specifying it twice (in the cwl and the call to picard, but any other way gets janky kind of fast


tmooney (approved):
+1 🧳

Issue #790, opened 2019-10-09
Split Interval List fails with `scatter_count` of 1.


tmooney:
When this value: 
https://github.com/genome/analysis-workflows/blob/dc2c019c1aa24cc01b451a0f048cf94a35f163c4/definitions/tools/split_interval_list.cwl#L20
is set to `1`, Picard's `SplitIntervalList` behaves differently and treats the value of `OUTPUT=` as a filename rather than a directory name.  This special-case behavior leads to workflow failures, since we specify an existing directory as the value in all cases.

jasonwalker80:
See #1002 for my first attempt at resolving this.

PR #1002, opened 2021-02-05
return the expected file path when a single scatter count is defined

jasonwalker80:
@tmooney Is this along the lines of what you were thinking? I could use some suggestion for testing. I started with a fresh docker image and I planned to add the script as a separate file with a test interval_list file. However, I realized I could simply switch from the cle image to a picard image and the perl snippet works inline.


jasonwalker80:
Closes #790

chrisamiller:
I'd have probably passed the inputs without the text and done something like this (rough and untested)


```
#!/bin/bash
set -o pipefail
set -o errexit
set -o nounset
            
SCATTERCOUNT=$1
INTERVALLIST=$2
OUTPUTDIR=$3

if [[ $SCATTERCOUNT == 1 ]];then
   cp $INTERVALLIST $OUTDIR
else
   /usr/bin/java -jar /usr/picard/picard.jar IntervalListTools SCATTER_COUNT=$SCATTERCOUNT
fi
```

You could even do that with your perl, but the point is that adding the `SCATTER_COUNT=` part in the script might save a bunch of complicated parsing

tmooney:
Chris has a good suggestion to take off the prefixes from the CWL.  Can then start the Perl script with something like:
```perl
die "wrong number of inputs" unless scalar(@ARGV) == 3;
my ($output_dir, $interval_list, $scatter_count) = @ARGV;
```
and remove a lot of the conditional logic in the new version of the script.

(Not that the `die` should be possible to hit when the CWL is working properly, but it could get tripped if the CWL is revised later.)

jasonwalker80:
Which position will the CWL `arguments` end up in the array? The two inputs are bound to specific positions. Will the "bare" argument be first or last?

chrisamiller:
it works like this:  
`-3   -2   -1   [arguments field]  1  2 3`

tmooney:
You already got an answer, but: arguments have a default position of `0` or you can also specify one explicitly if you ever need to, for example:
https://github.com/genome/analysis-workflows/blob/ab76e5d891b26fcc3c48a97aed55355a53d038ea/definitions/tools/varscan_germline.cwl#L13-L15
Or, using the other YAML syntax seen in this tool:
```cwl
arguments:
    [{ valueFrom: $(runtime.outdir), position: 0 }] 
```

definitions/tools/split_interval_list.cwl
```diff
@@ -7,37 +7,42 @@ requirements:
...
+            my ($output_dir, $interval_list, $scatter_count) = @ARGV;
+            
             my $i = 1;
-            for(glob('*/scattered.interval_list')) {
-                #create unique names and relocate all the scattered intervals to a single directory
-                File::Copy::move($_, qq{$i.interval_list});
-                $i++
+
+            if ($scatter_count == 1) {
+                File::Copy::move($interval_list,qq{$i.interval_list});
```
tmooney, line 23: I don't think we can get away with using `move` when copying the original input file--which should be read-only.  We'll have to make an actual copy in this case.

definitions/tools/split_interval_list.cwl
```diff
@@ -7,37 +7,42 @@ requirements:
...
       - entryname: 'split_interval_list_helper.pl'
         entry: |
             use File::Copy;
 
-            my $retval = system('/usr/bin/java', '-jar', '/usr/picard/picard.jar', 'IntervalListTools', @ARGV);
-            exit $retval if $retval != 0;
-
+            die "wrong number of inputs" unless scalar(@ARGV) == 3;
+            my ($output_dir, $interval_list, $scatter_count) = @ARGV;
+
```
tmooney, line 19: Just some whitespace.
```suggestion

```

jasonwalker80:
whitespace addresed, tested with cwltool, test data added to example_data.

chrisamiller:
+1 - love that you created a test for this cwl alone - we should all strive to do this moving forward!

tmooney (approved):
+1, it'd be nice to devise a more systematic approach to testing in our workflows!

PR #1003, opened 2021-02-08
Update pvactools CWLs to version 2.0.0

susannasiebert:
This involved quite a few changes to the inputs since those changed in the new version. The VEP version was also updated to one that has the new Frameshift plugin, which is required for running pVACseq 2.0.

@johnegarza if you could have a look at the immuno.cwl to make sure I didn't miss anything, that would be great.

### DIFF
--- a/definitions/tools/pvacbind.cwl
+++ b/definitions/tools/pvacbind.cwl
@@ -13,14 +13,14 @@ arguments: [
     { valueFrom: " && ", shellQuote: false },
     "export", "TMPDIR=/tmp/pvacbind",
     { valueFrom: " && ", shellQuote: false },
-    "/opt/conda/bin/pvacbind", "run",
+    "/usr/local/bin/pvacbind", "run",
     "--iedb-install-directory", "/opt/iedb",
     { position: 5, valueFrom: $(runtime.outdir) },
 ]
 requirements:
     - class: ShellCommandRequirement
     - class: DockerRequirement
-      dockerPull: "griffithlab/pvactools:1.5.13"
+      dockerPull: "griffithlab/pvactools:2.0.0"
     - class: ResourceRequirement
       ramMin: 16000
       coresMin: $(inputs.n_threads)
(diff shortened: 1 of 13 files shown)


definitions/tools/pvacbind.cwl
```diff
@@ -44,16 +44,26 @@ inputs:
...
         inputBinding:
             position: 4
-    epitope_lengths:
+    epitope_lengths_class_i:
         type: int[]?
         inputBinding:
-            prefix: "-e"
+            prefix: "-e1"
+            itemSeparator: ','
+            separate: false
```
tmooney, line 52: With 
```cwl
separate: false
```
an input like
```yaml
epitope_lengths_class_i: [5,6]
```
would become
```bash
-e15,6,7
```
in the command line.  I see that `separate: false` is in the old version, but should we take it out for this (and the class_ii input below it)?  (Repeat this comment for other tools with these inputs, as well.)

definitions/tools/pvacfuse.cwl
```diff
@@ -47,16 +47,26 @@ inputs:
...
         type: int[]?
         inputBinding:
-            prefix: "-e"
+            prefix: "-e1"
+            itemSeparator: ','
+            separate: false
+    epitope_lengths_class_ii:
+        type: int[]?
+        inputBinding:
+            prefix: "-e1"
```
tmooney, line 59: Looks like a typo here.
```suggestion
            prefix: "-e2"
```

susannasiebert: Fixed.

definitions/tools/pvacbind.cwl
susannasiebert: I updated all the tool CWLs to remove the separate option

tmooney (dismissed):
+1 🦝

johnegarza (approved):

johnegarza:
+1 immuno wf looks good!

gschang:
+1 These updates with new pVACtools 2.0.0 version are exciting. Jasreet, Mike, and I are discussing to use this version for the clinical studies. Nice presentation today.

PR #1008, opened 2021-02-15
Example YAML: bqsr_known_sites is an array of Files.

tmooney:
To match the type in pipelines/somatic_exome.cwl, this should be an array of `File` rather than of strings.

https://github.com/genome/analysis-workflows/blob/389f6edccab082d947bee9c032f59dbdf9f7c325/definitions/pipelines/somatic_exome.cwl#L69-L70


chrisamiller (approved):
+1  with the assumption that you've tested the example workflow and it runs with these changes!

PR #1007, opened 2021-02-12
Fix the problem of indexing an unsorted merged BAM in rnaseq.cwl

gschang:
**Background**:
There is a broken step in `definitions/pipelines/rnaseq.cwl` because it tries to index an unsorted BAM. For fixing this problem, a new step was added in order to sort a merged BAMs between the `merge` and `index_bam` step in rnaseq.cwl.

**Changes**:
There is a built-in cwl tool `definitions/tools/position_sort.cwl` for this purpose. I added it in rnaseq.cwl, and I changed inputs in the `index_bam` and `mark_up` step accordingly.

Issue #958, opened 2020-10-06
Validate Workflow Examples with cwltool


tmooney:
I was working on another issue and discovered that the example input YAML I was using was outdated and the workflow I was running could be run with Cromwell but not `cwltool`.  Before putting out a new release, it'd be good to polish up the examples and make sure they can be run with `cwltool`.  (Our current automated tests are pinned to an older version; should we also test against the latest `cwltool` for the release?)

This is somewhat related to #645, but with a different objective.

tmooney:
To clarify, this isn't only about `cwltool --validate`--there are some things that only come up at runtime.  (A bonus if we get some examples through with `cwltool` is that they could be useful for #116!)

gschang:
I started working from the latest immuno workflow.

gschang:
As new updates on testing with somatic_exome workflow examples, I finally got a succeeded build using the latest analysis-workflows Git repository (as of today, #1000 merged). My test job went smooth and I don't have any issue. All examples for somatic_exome workflow successfully work. Below are two points to discuss.


(1) Interpretation of results
I got only one somatic variant on TP53 as a result (ENSP00000269305.4:p.Arg175His missense mutation) from the current example sequencing data.

Hi, @chrisamiller. May I have your advice on this result interpretation? Do these results make sense to you? If you want to take a look at them, below is the results directory. Let me know if you cannot access that folder.

/gscuser/gchang/work2/14_BGA-WALKER/210203-cwltool2examples-958/1.analysis-workflows-test/example_data/test/results


(2) Next plan for aggregation

This test was done with my somatic_exome YAML update. But I didn't push it yet because I saw Chris's work on this in parallel--for example, VCFs for bqsr_known_sites and upgrade for per_target_intervals. I want to aggregate all these efforts to a final one.

Hi, @chrisamiller. Please let me know your thought and plan on somatic_exome test. I will find out what I can do more to finalize this somatic_exome workflow example. I would like to move on for other workflow examples--for example, rnaseq or bisulfite workflow (if you suggest). This is very exciting for me to get into and contribute to analysis workflows as Jason advised. I am happy to further discuss with you all.

gschang:
Here is a progress update on Workflow Examples, especially somatic_exome CWL workflow. As updated above, I confirmed that the current examples work without any problem for the latest somatic_exome workflow (at analysis-workflows Git).

[...]

PR #1013, opened 2021-03-03
Fix cgpbigwig_bamcoverage step and update workflow example of rnaseq.cwl

gschang:
This PR has two updates. I finally confirmed that the current rnaseq.cwl is now working with these updates.

(1) Fix `definitions/tools/bam_to_bigwig.cwl` as a subsequent update of #1007

The current rnaseq.cwl was broken in the `cgpbigwig_bamcoverage` step, because of invalid `MarkedSorted.bai` requirement. This PR fixes this broken step.

(2) Update rnaseq.cwl workflow example for the issue #958 

The previous workflow example was outdated. I updated the inputs YAML file, and I confirmed that this workflow example for `rnaseq.cwl` is perfectly working now.

### DIFF
--- a/definitions/tools/bam_to_bigwig.cwl
+++ b/definitions/tools/bam_to_bigwig.cwl
@@ -3,7 +3,7 @@
 cwlVersion: v1.0
 class: CommandLineTool
 label: 'cgpBigWig Converting BAM to BigWig'
-baseCommand: ["bam2bw", "-a -F", "1024"]
+baseCommand: ["bam2bw", "-a", "-F", "1024"]
 arguments: ["-o", { valueFrom: $(inputs.bam.nameroot).bw}]
 requirements:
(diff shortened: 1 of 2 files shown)


gschang:
@sridhar0605, could you take a look at this PR? As we discussed, I fixed the broken *rnaseq.cwl* workflow, and it is now working with the updated workflow example.

chrisamiller (approved):
Thanks for the fix.  +1

For this PR, the paths to `gscmnt` paths are fine, but we should get these pointed to things in the example_data files in a future PR

gschang:
Thank you for this review, Chris. I got your point to use built-in example datasets in the future, but not from `gscmnt`. I am going to merge it, once I discuss with @sridhar0605. He is working on for the rnaseq workflow and I am harmonizing with his work.

PR #1015, opened 2021-03-04
Switch to docker image hosted in our mgibio namespace.

### LINKED ISSUE #1011: strandedeness_check tool shouldn't use an ENTRYPOINT in its docker image
The image listed here:
https://github.com/genome/analysis-workflows/blob/93656ed6582073e434eab168c610625a835dce37/definitions/tools/strandedness_check.cwl#L28
uses an `ENTRYPOINT` specification, but Cromwell requires being able to run `/bin/bash` itself and thus the workflow will fail.

(On a certain internal cluster, the newer docker wrapper for LSF automatically clears all ENTRYPOINTs, so testing that way won't uncover this issue. Use `docker0` instead to see it!)

(Ideally we'd point to an image in our `genome` organization, too!)

tmooney:
Closes #1011.

### DIFF
--- a/definitions/tools/strandedness_check.cwl
+++ b/definitions/tools/strandedness_check.cwl
@@ -25,7 +25,7 @@ requirements:
       tmpdirMin: 5000
       coresMin: 1
     - class: DockerRequirement
-      dockerPull: "smk5g5/checkstranded:version1.0"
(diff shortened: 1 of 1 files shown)


chrisamiller (approved):
+1

PR #1017, opened 2021-03-11
Updated stringtie version to 2.1.4

smk5g5:
Updated version of stringtie from version 1.3.3 to 2.1.4.

Coverage,FPKM and TPM values do change in the stringtie_gene_expression.tsv output file for version 2.1.4 compared to version 1.3.3 but the workflow finishes successfully!


chrisamiller (approved):
+1 looks good to me!

PR #1019, opened 2021-03-23
Adding LLR and MAPQ column to variants table

saimukund20:


chrisamiller:
Sai, once you confirm that this runs the full somatic exome test without breaking, then happy to approve!

tmooney:
Something to be aware of: every `pipeline` and `subworkflow` that calls this tool defines its own `default` for this inputs, which will override the one here.  It's fine if we want to change the default when the tool is called directly, but I'm curious if any of the workflows should also have their defaults updated.

chrisamiller:
Good point. Sai, you probably want to be changing the defaults on the somatic WGS and somatic exome workflows, not here.

PR #1020, opened 2021-03-26
update mutect regex to better handle sample name extraction

chrisamiller:
The previous regex fails when there are lines in the bam header like this, which contain the SM tag as well as trailing information, all of which gets captured. 

`@PG     ID:bwa.2        PN:bwa  VN:0.7.12-r1039 CL:bwa mem -t 8 -T 0 -R @RG\tCN:BI\tDT:2010-03-12T00\tID:61D5N.8\tLB:Solexa-21878\tPL:illumina\tPU:61D5NAAXX100312.8\tSM:TCGA-AB-2821-03B-01W-0728-08 /alignment/reference/GRCh38.d1.vd1.fa /alignment/scratch1yVgmO/fastq/rmdup/61D5N.8_1.fq.gz /alignment/scratch1yVgmO/fastq/rmdup/61D5N.8_2.fq.gz`

This change restricts to looking at `@RG` lines and only captures the sample name. 

It will fail on sample names with spaces in them, but [a space in a filename is a space in one's soul](https://twitter.com/aaronquinlan/status/711593127551733761). (and I'm fairly confident that there are other places in these pipelines that would break with spaces as well)

### DIFF
--- a/definitions/tools/mutect.cwl
+++ b/definitions/tools/mutect.cwl
@@ -22,8 +22,8 @@ requirements:
             export tumor_bam="$3"
             export normal_bam="$4"
 
-            TUMOR=`perl -e 'my $header_str = qx(samtools view -H $ENV{tumor_bam}); my ($sample_name) = $header_str =~ /@RG.+SM:([^\s]+)/; print $sample_name'` #Extracting the sample name from the TUMOR bam.
-            NORMAL=`perl -e 'my $header_str = qx(samtools view -H $ENV{normal_bam}); my ($sample_name) = $header_str =~ /@RG.+SM:([^\s]+)/; print $sample_name'` #Extracting the sample name from the NORMAL bam.
+            TUMOR=`perl -e 'my $header_str = qx(samtools view -H $ENV{tumor_bam}); my ($sample_name) = $header_str =~ /^@RG.+\tSM:([ -~]+)/; print $sample_name'` #Extracting the sample name from the TUMOR bam.
+            NORMAL=`perl -e 'my $header_str = qx(samtools view -H $ENV{normal_bam}); my ($sample_name) = $header_str =~ /^@RG.+\tSM:([ -~]+)/; print $sample_name'` #Extracting the sample name from the NORMAL bam.
             /gatk/gatk Mutect2 --java-options "-Xmx20g" -O $1 -R $2 -I $3 -tumor "$TUMOR" -I $4 -normal "$NORMAL" -L $5 #Running Mutect2.
             /gatk/gatk FilterMutectCalls -R $2 -V mutect.vcf.gz -O mutect.filtered.vcf.gz #Running FilterMutectCalls on the output vcf.
 
--- a/definitions/tools/mutect.cwl
+++ b/definitions/tools/mutect.cwl
@@ -22,8 +22,8 @@ requirements:
             export tumor_bam="$3"
             export normal_bam="$4"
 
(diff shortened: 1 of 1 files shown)


tmooney:
I'm a little confused by this change.  The previous regex is taken directly from 1.3 of the [SAM specification](https://samtools.github.io/hts-specs/SAMv1.pdf).  That makes me think this is an invalid case this change is trying to handle?  (It's not quite clear to me what's a tab vs a space vs a newline in that example header.)

chrisamiller:
Here's a little more context from this header, which is taken straight from the NCI genomic data commons. 

```
@PG                 ID:bwa              PN:bwa              VN:0.7.12-r1039     CL:bwa mem -t 8 -T 0 -R @RG\tCN:BI\tDT:2010-03-12T00\tID:61D5N.4\tLB:Solexa-21878\tPL:illumina\tPU:61D5NAAXX100312.4\tSM:TCGA-AB-2821-03B-01W-0728-08 /alignment/reference/GRCh38.d1.vd1.fa /alignment/scratch1yVgmO/fastq/rmdup/61D5N.4_1.fq.gz /alignment/scratch1yVgmO/fastq/rmdup/61D5N.4_2.fq.gz
@PG                 ID:bwa.1            PN:bwa              VN:0.7.12-r1039     CL:bwa mem -t 8 -T 0 -R @RG\tCN:BI\tDT:2010-03-12T00\tID:61D5N.6\tLB:Solexa-21878\tPL:illumina\tPU:61D5NAAXX100312.6\tSM:TCGA-AB-2821-03B-01W-0728-08 /alignment/reference/GRCh38.d1.vd1.fa /alignment/scratch1yVgmO/fastq/rmdup/61D5N.6_1.fq.gz /alignment/scratch1yVgmO/fastq/rmdup/61D5N.6_2.fq.gz
@RG                 ID:61D5N.4          CN:BI               DT:2010-03-12T00:00:00+0000             LB:Solexa-21878     PL:illumina         PU:61D5NAAXX100312.4                    SM:TCGA-AB-2821-03B-01W-0728-08
@RG                 ID:61D5N.6          CN:BI               DT:2010-03-12T00:00:00+0000             LB:Solexa-21878     PL:illumina         PU:61D5NAAXX100312.6                    SM:TCGA-AB-2821-03B-01W-0728-08

```
The regex as exists matches `SM:...` multiple times in this header.  It matches on the @RG lines, which is expected, but also matches the command line string, where fields were specified explicitly as inputs with `\t` (not tab characters), thus giving the first result as:

```
TCGA-AB-2821-03B-01W-0728-08 /alignment/reference/GRCh38.d1.vd1.fa /alignment/scratch1yVgmO/fastq/rmdup/61D5N.8_1.fq.gz /alignment/scratch1yVgmO/fastq/rmdup/61D5N.8_2.fq.gz
```
(matches the SM field, plus everything folllowing it)

After review, you're right that we could also solve this by restricting the existing code to only look at `@RG` lines (I misunderstood how the pattern matching was working in that regex).  I'll make that modification and test it.

tmooney:
Thanks for the explanation!  Including the tab in the regex preceding the `SM:` could solve this (since the command line must necessarily escape it). Restricting to `@RG` lines seems like a good idea, too.

chrisamiller:
I was going to go with `/^@RG.+SM:([ -~]+)/`, but do I understand right that you're suggesting `/^@RG.+\tSM:([ -~]+)/`?

tmooney:
Yeah, I'd include a `\t`.  (I might hope nobody makes a library name with `SM:` in it, but y'never know 😄)

chrisamiller:
modified and tested - ready for re-review!

tmooney (approved):
+1 🧷

PR #1021, opened 2021-03-29
updating regex to handle edge case headers

chrisamiller:
Found another place where this sample extraction code breaks (see https://github.com/genome/analysis-workflows/pull/1020).  Did a quick grep through the repo and don't see any others at present.


tmooney (approved):
+1 ⚓

PR #1022, opened 2021-03-31
updated hisat to version 2.2.1

smk5g5:
updated hisat to version 2.2.1


chrisamiller (approved):
+1   Assuming you've tested it on the example data, it looks good to me!

PR #1024, opened 2021-04-19
wgs_cnvkit

bryanfisk:
Adds cnvkit step to somatic_wgs pipeline and changes how sample/normal_sample_names are handled in the bam_readcount and add...bam_readcount_to_tsv steps of detect_variants_wgs.cwl.


definitions/pipelines/somatic_wgs.cwl
```diff
@@ -166,6 +166,9 @@ inputs:
         type: File?
         secondaryFiles: [.tbi]
         doc: "An optional VCF with variants that will be flagged as 'VALIDATED' if found in this pipeline's main output VCF"
+    cnvkit_target_average_size:
+        type: int?
+        doc: "approximate size of split target bins for CNVkit"
```
chrisamiller, line 171: Is there a sensible default already set here in the tool? (it's not set in the CWL explicitly)  If so, might be worth mentioning that in the doc line here.  ("if not provided, CNVkit will use XX")

chrisamiller (approved):
+1 looks good!

PR #1027, opened 2021-05-05
update bcftools version 1.12

apaul7:
updating bcftools version for #948
version 1.9 to 1.12, release notes can be found [here](https://github.com/samtools/bcftools/releases)
One new feature is that the output type no longer needs to be specified. bcftools is able to determine the output type based on the suffix of the output file name.


tmooney (approved):
+1, a few of these are updating from version 1.3.1 which is quite a bit older!  I am trusting the new version is still compatible inasmuch as we care.

PR #1031, opened 2021-05-19
Added CNVkit to somatic exome nonhuman and cnvkit_target_average_size to somatic exome

bryanfisk:


chrisamiller (approved):
+1 - I don't see any issues here. 

One question - is there a small test data that you used for testing here?  We'd like to get the example data directory to have more comprehensive coverage (if not, we can cross that bridge later).

tmooney:
Are the changes to `detect_variants_wgs.cwl` meant to be included in this PR?  They don't really match up with the included description at all.

bryanfisk:
Well the pipeline would fail until I made those changes but, no, they don't really have much else to do with the CNVkit part.

tmooney:
Yeah, the changes seem like they make sense!  I just didn't see a connection to CNVkit. Ideally if they're not directly related they'd have their own commit or PR (or if nothing else get mentioned in the text of the PR).

PR #1037, opened 2021-07-06
force germline workflows to produce genotyped vcf

apaul7:
This PR removes the `emit_reference_confidence` input for germline workflows, defaulting it to `NONE`. If the input value was not set to `NONE` the output vcf would not be a genotyped VCF but a genome VCF(gVCF). This causes issues with downstream steps. 

@dufeiyu, should I also change the `pipelines/aml_trio_cle.cwl` and `pipelines/amle_trio_cle_gathered.cwl` workflows?


apaul7:
We have the `pipelines/germline_exome_gvcf.cwl` and `pipelines/germline_wgs_gvcf.cwl` workflows. They can produce the gvcf files. They run the alignment workflow and produce gvcf files. They do not make a genotyped vcf. 

Currently, we do not have a workflow that can produce gvcf files and a final genotyped vcf.

tmooney (approved):
+1 🚀
Certainly making a version of the pipeline that creates both gvcf and a final vcf is a bigger change than this PR needs to be worrying about! 😄

jasonwalker80 (approved):
+1

PR #1039, opened 2021-07-12
Initial github action for testing

tmooney:
Since travis-ci.org is gone, we'll need a replacement.  This tries to faithfully reproduce the existing .travis.yml as a github workflow.  (If this works something similar needs to happen for the auto-documentation, too.


apaul7 (approved):
+1

PR #1042, opened 2021-07-14
Set user name and email for documentation update commit.

tmooney:
On Travis the default info was enough to generate a commit, but github doesn't fill in enough info in its actions, so let's set up a name and email for ourselves!


apaul7 (approved):
+1

PR #1043, opened 2021-07-15
germline joint detect variants workflow

apaul7:
This PR adds a joint detect variants subworkflow. This is split into 2 separate subworkflows, joint_detect_snps.cwl and joint_detect_svs.cwl. Tried my best to add informative commit messages. I've itemized most of the overall changes below:

- `tools/replace_vcf_sample_name.cwl`
  update naming of output file to full name instead of just the base with renamed in front
- `tools/gather_to_subdirectory.cwl`
  added --recursive, --preserve, and --no-clobber to the cp command. allows timestamps to be preserved and errors thrown if files are overwritten
  also added the valueFrom field which uses javascript to iterate over the input array and add any secondaryFiles to the destination directory
- `tools/gather_to_subdirectory_dirs.cwl`(new tool)
  this follows the same format as the tools/gather_to_subdirectory.cwl format but is used for directories. I tried using the same tool for both use cases but was unable to make Cromwell happy
- `tools/bcftools_view.cwl`(new tool)
  This tool is used to split multi sample vcfs into single sample vcfs.
- `tools/vt_normalize.cwl`(new tool)
  uses VT to normalize a vcf, alternative to gatk LeftAlignAndTrimVariants found in `tools/normalize_variants.cwl`
- `subworkflows/joint_genotype.cwl`
  added decompose and normalize steps
- `tools/manta_germline.cwl`(new tool)
  follows `tools/manta_somatic.cwl` format. adds the stats directory to outputs and removes somatic and tumor only outputs.
- `tools/genotype_gvcfs.cwl`
  added input for minimum confidence threshold for called variants
- `tools/custom_merge_sv_records.cwl`(new tool)
  merges copy number called variants that have the same type, and are within x bases
- `tools/cnvnator.cwl`
  updated output file names, s/CNV/cnv/
- `tools/annotsv_filter.cwl`
  adds ability to merge survivor merged vcf, skips last filtering requirement. survivor does not pass INFO fields to merged vcf
  renamed all_cds to no_cds. easier to understand that input removes the coding sequence filter requirement
- `tools/annotsv.cwl`
  updated to version 2.3. This is not the latest version. The latest version no longer retains information for individual sv population databases in output files
  renamed inputs for the new version
  added input for annotation directory instead of having them in the docker image
  added unannotated tsv output
- `subworkflows/merge_svs.cwl`
  added inputs for population allele frequency, no_cds annotsv filtering, and anntosv annotation directory
  output file name replacement, s/SURVIVOR/survivor/
  added step for survivor merged annotsv filtering
- `subworkflows/gatk_soft_filter.cwl`
  added subworkflow for gatk soft filter based on hard parameters to add PASS/FAIL 
  https://gatk.broadinstitute.org/hc/en-us/articles/360035531112--How-to-Filter-variants-either-with-VQSR-or-by-hard-filtering#2

### DIFF
--- a/definitions/tools/bcftools_view.cwl
+++ b/definitions/tools/bcftools_view.cwl
@@ -1,0 +1,53 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+
+baseCommand: ["/opt/bcftools/bin/bcftools", "view"]
+
+requirements:
+    - class: ResourceRequirement
+      ramMin: 4000
+    - class: DockerRequirement
+      dockerPull: "mgibio/bcftools-cwl:1.12"
+
+inputs:
+    sample_name:
+        type: string?
+        inputBinding:
+            position: 1
+            prefix: "--samples"
+        doc: "comma separated list of samples to include (or exclude with '^' prefix)"
+    output_type:
+        type:
+            type: enum
+            symbols: ["b", "u", "z", "v"]
+        default: "z"
+        inputBinding:
+            position: 4
+            prefix: "--output-type"
+        doc: "output file format"
+    output_vcf_name:
+        type: string?
+        default: "bcftools_split.vcf.gz"
+        inputBinding:
+            position: 5
+            prefix: "--output-file"
+        doc: "output vcf file name"
+    variant_type:
+        type: string?
+        inputBinding:
+             position: 6
+             prefix: "--types"
+        doc: "select comma-separated list of variant types: snps,indels,mnps,ref,bnd,other"
+    in_vcf:
+        type: File
+        inputBinding:
+            position: 7
+        doc: "input bgzipped tabix indexed vcf to view"
+
+outputs:
+    vcf:
+        type: File
+        outputBinding:
+            glob: $(inputs.output_vcf_name)
--- a/definitions/tools/custom_merge_sv_records.cwl
+++ b/definitions/tools/custom_merge_sv_records.cwl
@@ -1,0 +1,49 @@
+#!/usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+label: "merges nearby DEL/DUP records within a certain window distance"
+
+baseCommand: ["/bin/bash", "run_merge.sh"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 4000
+    - class: DockerRequirement
+      dockerPull: "apaul7/analysis:1.0.0"
+    - class: InitialWorkDirRequirement
+      listing:
+      - entryname: "run_merge.sh"
+        entry: |
+          #!/bin/bash
+          set -eou pipefail
+          INPUT="$1"
+          OUTPUT="$2"
+          DISTANCE="$3"
+          /usr/local/bin/python3 /opt/git/merge-sv-records/merge.py -i $INPUT -o $OUTPUT -w $DISTANCE
+
+          /usr/local/bin/bgzip $OUTPUT
+          /usr/local/bin/tabix -p vcf $OUTPUT.gz
+
+
+inputs:
+    input_vcf:
+        type: File
+        inputBinding:
+            position: 1
+    output_vcf_name:
+        type: string?
+        default: "record_merged.vcf"
+        inputBinding:
+            position: 2
+    distance:
+        type: int?
+        default: 1000
+        inputBinding:
+            position: 3
+
+outputs:
+    vcf:
+        type: File
+        outputBinding:
+            glob: "$(inputs.output_vcf_name).gz"
+        secondaryFiles: [.tbi]
--- a/definitions/tools/gather_to_sub_directory_dirs.cwl
+++ b/definitions/tools/gather_to_sub_directory_dirs.cwl
@@ -1,0 +1,40 @@
+#! /usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+baseCommand: ["/bin/bash","directory_gatherer.sh"]
+
+requirements:
+    - class: DockerRequirement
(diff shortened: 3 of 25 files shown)


jasonwalker80 (dismissed):
+1, @apaul7 I have not reviewed this in it's entirety. I've asked @tmooney to take a look as well. Both of us have reviewed the PR but not necessarily commit-by-commit or line-by-line. I'm going to give the "looks good to me", but if Tom can look mostly for places where these commits/changes may or may not impact other workflows/pipelines. Then let's merge.

definitions/tools/bcftools_view.cwl
```diff
@@ -0,0 +1,53 @@
...
+    - class: DockerRequirement
+      dockerPull: "mgibio/bcftools-cwl:1.12"
+
+inputs:
+    sample_name:
+        type: string?
+        inputBinding:
+            position: 1
+            prefix: "--samples"
+        doc: "comma separated list of samples to include (or exclude with '^' prefix)"
```
tmooney, line 20:
```suggestion
        doc: "comma-separated list of samples to include (or exclude with '^' prefix)"
```

definitions/tools/bcftools_view.cwl
```diff
@@ -0,0 +1,53 @@
...
+    output_type:
+        type:
+            type: enum
+            symbols: ["b", "u", "z", "v"]
+        default: "z"
+        inputBinding:
+            position: 4
+            prefix: "--output-type"
+        doc: "output file format"
```
tmooney, line 29: Is it worth mentioning what the options mean?

definitions/tools/custom_merge_sv_records.cwl
```diff
@@ -0,0 +1,49 @@
...
+cwlVersion: v1.0
+class: CommandLineTool
+label: "merges nearby DEL/DUP records within a certain window distance"
+
+baseCommand: ["/bin/bash", "run_merge.sh"]
+requirements:
+    - class: ResourceRequirement
+      ramMin: 4000
+    - class: DockerRequirement
+      dockerPull: "apaul7/analysis:1.0.0"
```
tmooney, line 12: Is this an image that we should promote to the `mgibio` namespace?

definitions/tools/custom_merge_sv_records.cwl
```diff
@@ -0,0 +1,49 @@
...
+        entry: |
+          #!/bin/bash
+          set -eou pipefail
+          INPUT="$1"
+          OUTPUT="$2"
+          DISTANCE="$3"
+          /usr/local/bin/python3 /opt/git/merge-sv-records/merge.py -i $INPUT -o $OUTPUT -w $DISTANCE
+
+          /usr/local/bin/bgzip $OUTPUT
+          /usr/local/bin/tabix -p vcf $OUTPUT.gz
```
tmooney, line 25: Maybe we should quote the parameters with variables on these lines just in case the names have some weird characters in them.

definitions/tools/gather_to_sub_directory_dirs.cwl
```diff
@@ -0,0 +1,40 @@
+#! /usr/bin/env cwl-runner
+
+cwlVersion: v1.0
+class: CommandLineTool
+baseCommand: ["/bin/bash","directory_gatherer.sh"]
+
+requirements:
+    - class: DockerRequirement
+      dockerPull: "ubuntu:xenial"
```
tmooney, line 9: Is it time to start basing things on `bionic` or even `focal`?

definitions/tools/gather_to_sub_directory_dirs.cwl
```diff
@@ -0,0 +1,40 @@
...
+      listing:
+      - entryname: 'directory_gatherer.sh'
+        entry: |
+            set -eou pipefail
+
+            outdir="$1"
+            files="${@:2}"
+            mkdir $outdir
+            chmod -R 777 $outdir
+            cp --recursive --preserve --no-clobber --target-directory $outdir $files
```
tmooney, line 22: I'll advocate for quoting the parameters here, too.

tmooney (commented):
I generally think this looks okay.  I'll trust that what it does is what you want to have happen.

Part of me sees all the references to SNPs has flashbacks to past instances where we've been told to go back and change it to SNVs instead, but I'll also trust that this name is the one we want to use 😄

definitions/tools/bcftools_view.cwl
apaul7: done!

definitions/tools/custom_merge_sv_records.cwl
apaul7: I moved the script in the cwl and used a different docker image

definitions/tools/custom_merge_sv_records.cwl
apaul7: I changed to a python script for this tool

definitions/tools/gather_to_sub_directory_dirs.cwl
apaul7: I changed to the `focal` image

definitions/tools/gather_to_sub_directory_dirs.cwl
apaul7: Quoting the `files` variable during the `cp` command has the command think it's a single file with spaces instead of multiple space separated files. This causes the tool to fail unless there is a single input file. I kept thinking it was a compute cache layer issue. Have any other suggestions?

tmooney: The trick here is to keep the `$files` variable as a bash array, too.

So something like:
```bash
files=("${@:2}")
...
cp ... "${files[@]}"
```

apaul7: TIL! That method worked, thanks.

PR #1046, opened 2021-08-19
Update merge_bams.cwl to work with both cwltool and cromwell.

tmooney:
The `${...}` syntax is problematic, since cromwell doesn't respect the escaping described in the CWL specification.  Thus if you properly escape for `cwltool` the script is wrong in cromwell and if you don't escape it then `cwltool` complains about JavaScript issues.  This replaces the bash script with a perl script (`perl` 5.22 is already installed in this docker image.) to avoid needing the `${...}` syntax altogether.

(Also, for some reason the boolean default value was quoted here even though it wasn't other places. `cwltool` complains about that, too, so this PR removes the quotes.)

### DIFF
--- a/definitions/tools/merge_bams.cwl
+++ b/definitions/tools/merge_bams.cwl
@@ -66,7 +66,7 @@ inputs:
             position: 3
     sorted:
         type: boolean?
-        default: "false"
+        default: false
         inputBinding:
             prefix: "-s"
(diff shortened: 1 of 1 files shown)


apaul7 (approved):
+1

chrisamiller:
+1 nice fix

PR #1059, opened 2022-01-20
removing pindel param that isn't used

chrisamiller:
we dropped pindel from wgs pipelines a while back because of runtime


sridhar0605:
+1

tmooney (approved):
+1 makes sense 🪅

Not sure the file mode change is intentional here?  But in general I think it's a good idea, too.

PR #1060, opened 2022-03-03
handle unparseable GT entries by skipping those lines

johnmaruska:
Sometimes we'll encounter entries which have a $gt_str which has
characters not contained in the $ids map. This results in malformed gt
entries, like `/0` or `1/` or `/`. When these entries occur,
downstream calls like merge_vcf (in the case of immuno workflow) will
print an error and _silently continue on their merry way_. I have no
idea why they do this, but it's something we noticed when testing
conversion over to WDL with HCC1395 sample.

We address this issue by checking the results of @gt_ids, and if any
are undefined we eject early, returning undefined. At the call site we
handle this case separate by printing to STDERR, dropping that line,
and moving on to the remainder.

An alternative approach of a more strict failure is an option, but
discussion with Chris Miller and Malachi Griffith landed on decision
that we just drop and move on.

### DIFF
--- a/definitions/tools/add_strelka_gt.cwl
+++ b/definitions/tools/add_strelka_gt.cwl
@@ -55,11 +55,9 @@ requirements:
                             $id++;
                         }
 
-                        # parse GT, handling unparseable case with stderr
-                        my $parsed_n_gt = parse_gt($n_gt_str, \%ids);
                         if ($n_gt_info eq "ref") {
                             $n_gt = "0/0";
-                        } elsif (defined($parsed_n_gt)) {
+                        } elsif (defined(my $parsed_n_gt = parse_gt($n_gt_str, \%ids))) {
                             $n_gt = $parsed_n_gt;
                         } else {
                             my $id_keys = join(" ", sort keys(%ids));
@@ -67,8 +65,7 @@ requirements:
                             next;
                         }
 
-                        my $parsed_t_gt = parse_gt($t_gt_str, \%ids);
-                        if (defined($parsed_t_gt)) {
+                        if (defined(my $parsed_t_gt = parse_gt($t_gt_str, \%ids))) {
                             $t_gt = $parsed_t_gt;
                         } else {
                             my $id_keys = join(" ", sort keys(%ids));
--- a/definitions/tools/add_strelka_gt.cwl
+++ b/definitions/tools/add_strelka_gt.cwl
@@ -106,8 +106,7 @@ requirements:
             sub parse_gt {
                 my ($gt_str, $ids) = @_;
                 my @gt_ids = map{$ids->{$_}}(split //, $gt_str);
-                my $i;
-                foreach $i (@gt_ids) {
+                foreach my $i (@gt_ids) {
                     unless (defined $i) {
                         return undef;
                     }
--- a/definitions/tools/add_strelka_gt.cwl
+++ b/definitions/tools/add_strelka_gt.cwl
@@ -55,6 +55,27 @@ requirements:
                             $id++;
                         }
 
+                        # parse GT, handling unparseable case with stderr
+                        my $parsed_n_gt = parse_gt($n_gt_str, \%ids);
+                        if ($n_gt_info eq "ref") {
+                            $n_gt = "0/0";
+                        } elsif (defined($parsed_n_gt)) {
+                            $n_gt = $parsed_n_gt;
+                        } else {
+                            my $id_keys = join(" ", sort keys(%ids));
+                            print STDERR "parse_gt for n_gt=\"$n_gt_str\" and ids=\"$id_keys\" resulted in undefined\n";
+                            next;
+                        }
+
+                        my $parsed_t_gt = parse_gt($t_gt_str, \%ids);
+                        if (defined($parsed_t_gt)) {
+                            $t_gt = $parsed_t_gt;
+                        } else {
+                            my $id_keys = join(" ", sort keys(%ids));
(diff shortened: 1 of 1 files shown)


definitions/tools/add_strelka_gt.cwl
```diff
@@ -55,6 +55,27 @@ requirements:
...
 
+                        # parse GT, handling unparseable case with stderr
+                        my $parsed_n_gt = parse_gt($n_gt_str, \%ids);
+                        if ($n_gt_info eq "ref") {
+                            $n_gt = "0/0";
+                        } elsif (defined($parsed_n_gt)) {
+                            $n_gt = $parsed_n_gt;
+                        } else {
+                            my $id_keys = join(" ", sort keys(%ids));
+                            print STDERR "parse_gt for n_gt=\"$n_gt_str\" and ids=\"$id_keys\" resulted in undefined\n";
```
tmooney, line 66: Looks like other parts of this script are using `say` instead of `print` if you're interested in doing that here, too. I like the cleanliness of not needing to add `"\n"` manually with it.

definitions/tools/add_strelka_gt.cwl
```diff
@@ -88,6 +109,12 @@ requirements:
             sub parse_gt {
                 my ($gt_str, $ids) = @_;
                 my @gt_ids = map{$ids->{$_}}(split //, $gt_str);
+                my $i;
+                foreach $i (@gt_ids) {
```
tmooney, line 113: If you don't need it to persist outside of the loop, it might be better to declare the  loop variable directly in the for loop, like: 
```perl
for my $i (@gt_ids) {
```
(`for` and `foreach` are the same so whichever you prefer 😄)

definitions/tools/add_strelka_gt.cwl
```diff
@@ -55,6 +55,27 @@ requirements:
...
+                        my $parsed_t_gt = parse_gt($t_gt_str, \%ids);
+                        if (defined($parsed_t_gt)) {
+                            $t_gt = $parsed_t_gt;
+                        } else {
+                            my $id_keys = join(" ", sort keys(%ids));
+                            print STDERR "parse_gt for t_gt=\"$t_gt_str\" and ids=\"$id_keys\" resulted in undefined\n";
+                            next;
+                        }
+
                         $n_gt = $n_gt_info eq 'ref' ? '0/0' : parse_gt($n_gt_str, \%ids);
```
tmooney, line 79: Looks like if we get this far we're throwing away all the work in the new code and going back to the old way.

definitions/tools/add_strelka_gt.cwl
```diff
@@ -55,6 +55,27 @@ requirements:
                             $id++;
                         }
 
+                        # parse GT, handling unparseable case with stderr
+                        my $parsed_n_gt = parse_gt($n_gt_str, \%ids);
+                        if ($n_gt_info eq "ref") {
+                            $n_gt = "0/0";
```
tmooney, line 61: Since this case doesn't depend on parsing at all, maybe we should skip calling `parse_gt` if we're ending up here.

definitions/tools/add_strelka_gt.cwl
```diff
@@ -88,6 +109,12 @@ requirements:
             sub parse_gt {
                 my ($gt_str, $ids) = @_;
                 my @gt_ids = map{$ids->{$_}}(split //, $gt_str);
+                my $i;
+                foreach $i (@gt_ids) {
+                    if (!defined($i)) {
```
tmooney, line 114: This is completely fine; just noting that this is equivalent to:
```perl
unless (defined $i) {
```
in case you like having fewer operators and/or parentheses.

tmooney (commented):
This generally seems fine; most of my suggestions are entirely optional, but I think we should clean up the duplicated work before merging this.

definitions/tools/add_strelka_gt.cwl
johnmaruska: I had tried that earlier and for some reason it kept failing on me which is why I pulled it out. I can double back and try it again but that's how I landed here

definitions/tools/add_strelka_gt.cwl
johnmaruska: agh I forgot to delete the old lines when I copied the new ones over from the WDL I tested in. Fixing now

definitions/tools/add_strelka_gt.cwl
johnmaruska: I had that initially and I brought it outside mostly to avoid the extra nesting, especially since the strings are usually so short. I don't mind switching if others have a preference though

I'm guessing perl doesn't have the assignment expression thing python added in 3.8?
https://www.python.org/dev/peps/pep-0572/
I'd very much like to just have a single case that says "if parsed_n_gt is assigned with this call" which gives us both

tmooney: Are you asking if you can declare variables in the middle of expressions?  If so, yeah--you could do something like this if you wanted:

```perl
if (defined( my $parsed_n_gt = parse_gt(...) )) {
  say ("Got $parsed_n_gt as the value");
}
```
It'll be scoped to the conditional as if you did something like:
```perl
{
  my $parsed_n_gt = parse_gt(...);
  if (defined($parsed_n_gt)) {
     say ("Got $parsed_n_gt as the value");
  }
}
```
which is also to say you should be able to then use the value inside the blocks following `elsif` or `else` if needed, but not beyond.

tmooney (approved):
+1 makes sense to me!

PR #1063, opened 2022-04-06
Patch case where clinical class II alleles are not provided

johnegarza:


susannasiebert (approved):

gschang:
(+1) I agree that this patch solves this issue.

PR #1064, opened 2022-04-19
fix docm filter by passing decomposed vcf in tumor only mode

### LINKED ISSUE #1062: germline docm filter needs a decompose step
Add a decompose step to the VCF file (as is done in the docm_cle somatic workflow), so that the perl script filters correctly.  As is, it doesn't handle multi-allelic lines correctly.

Site to insert:
https://github.com/genome/analysis-workflows/blob/fbeea265295ae596d5a3ba563e766be0c4fc26e8/definitions/subworkflows/docm_germline.cwl#L40

Decompose example:
https://github.com/genome/analysis-workflows/blob/fbeea265295ae596d5a3ba563e766be0c4fc26e8/definitions/subworkflows/docm_cle.cwl#L55

sridhar0605:
Fixes #1062


chrisamiller (approved):
+1

PR #1069, opened 2022-05-24
allow optional normal input in mutect tool

apaul7:
The mutect tool can be ran in tumor only mode. The cwl also allows the normal bam to be an optional input. The bash script does not. This PR adds an if/else statement to the bash script to allow the tool to successfully run when the normal bam is not provided.


tmooney (dismissed):
+1, I'd be tempted to add `-o nounset` as long as we're modifying this script.

apaul7:
good idea @tmooney!

tmooney (approved):
+1 again 🪧

PR #1071, opened 2022-06-02
actually allow nounset for tumor only mutect runs

apaul7:
having the `nounset` option set does not allow mutect to run in tumor only mode. The normal bam is an optional input and the bash script exits before running when the normal bam is not provided.


chrisamiller (approved):
+1

tmooney:
Could also choose to only assign `$5` inside the case where we have it.

PR #1073, opened 2022-08-09
Upgrade mapq0 filter to v0.5.3

tmooney:
This handles variants with no DP information.  (For more details, see genome/docker-mapq0-filter#8.)


chrisamiller (approved):
+1

PR #1077, opened 2022-09-13
Fda metrics split on sequence

johnegarza:
This PR introduces a CWL scatter across individual sequence elements (a bam or pair of fastqs) of the input array for each of the 3 unaligned inputs (normal DNA, tumor DNA, tumor RNA) when calculating statistics to be reported to the FDA in clinical cases. Previously, some statistics were calculated using data from all sequence files for an input at once, which is less accurate.

### DIFF
--- a/definitions/tools/unaligned_seq_fda_stats.cwl
+++ b/definitions/tools/unaligned_seq_fda_stats.cwl
@@ -14,18 +14,6 @@ requirements:
         entry: |
                 #!/usr/bin/perl
 
-                # It is free software; you can redistribute it and/or modify it under the terms of
-                # the GNU General Public License (GPLv3) as published by the Free Software Foundation.
-                #
-                # McDonnell Genome Institute
-                # Washington University School of Medicine
-                # 4444 Forest Park Ave
-                # St. Louis, Missouri 63108
-                # 
-                # http://genome.wustl.edu
-
-
-                # includes perl modules
                 use strict;
                 use warnings;
                 use Carp;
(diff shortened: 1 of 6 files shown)


gschang:
Thank you for this work, John. These updates are about FDA QC metric reports, which John and we recently added to the immuno workflow. We discussed more consistent QC reporting with multiple instrument data, i.e. towards more generalized interface to include them in the report.

As John suggests, I plan to run a test to doublecheck the key changes in the final FDA QC metrics report.

definitions/tools/unaligned_seq_fda_stats.cwl
```diff
@@ -121,6 +117,22 @@ requirements:
                 sub Main {
                     # local variables
                     
+                    
+                    # gets the input file format
+                    unless (defined $format)
```
tmooney, line 122: It looks like this is being defined on line 72.  Will we ever fall into this case?  If we don't fall into this case for whatever reason, does that mean the user must NOT pass the format lest it be added to `@paths` instead?

definitions/tools/unaligned_seq_fda_stats.cwl
```diff
@@ -251,6 +262,15 @@ requirements:
                     }
                     
                     
+                    # prints out the source file information
+                    printf "\n\n[Input file information]";
+                    foreach my $path (@paths)
+                    {
+                        my ($name, $dir, $ext) = fileparse($path, qr/\.[^.]*/);
+                        printf "\n%s\t%s", $name . $ext, $dir;
```
tmooney, line 270: If you don't care about the extension, can simplify this:
```suggestion
                        my ($name, $dir) = fileparse($path);
                        printf "\n%s\t%s", $name, $dir;
```

tmooney (commented):
In general I'm a little puzzled by some of the commented-out code and related conditionals in these scripts.  Seems like it makes the logic more complicated and invites the potential for incorrect behavior if the wrong combination of lines are uncommented.

definitions/tools/unaligned_seq_fda_stats.cwl
gschang: These comments are good points. I am going to clean them up in the source codes as Tom suggests, and also removing comment lines that are not in the right place.

gschang:
I just want to update this ticket here and John, Jasreet, and I are working on this PR. John and we have discussed more upgrades for generalized workflow concept, and now in progress. I updates the scripts as Tom advised above and finished testing.

gschang:
These new changes in the `generate_fda_metrics.cwl` subworkflow worked on my end. I doublechecked every upgrade/fixation.

We're going to release a new immuno workflow unless there are other comments from someone else. I have requested John to create a new immuno processing profile in two versions--(1) immuno.cwl (currently running) and (2) immuno.cwl without rnaseq input.

gschang (approved):

PR #1079, opened 2022-10-18
FDA table generation: the output `table_file_name` is required

tmooney:
The python script requires this to be specified (and currently all callers do specify it), so let's make it required to help prevent any future mistakes.


gschang:
(+1)

PR #1082, opened 2022-12-13
Use sambamba for duplicate marking.

tmooney:
In porting the FDA metrics pipelines to the WDL workflows I ran into that those had been updated to switch duplicate markers but these had not.  This carries that change over to the CWL version.  I haven't tried running it on the immuno.cwl yet--thought I'd open this and see if it's something we wanted to do before running a test.


chrisamiller:
I'll let John and others review, but I support this change as one that increases the speed through parallelization while using the same duplicate marking strategy and producing essentially identical results.

PR #1085, opened 2023-03-10
Updates to immuno workflow for upcoming 4.0.0 release

susannasiebert:
This implements the features as discussed here: https://github.com/wustl-oncology/analysis-wdls/issues/87

Still outstanding is the addition of Arriba to run in parallel to starfusion/AGfusion.

### DIFF
--- a/definitions/tools/pvacseq.cwl
+++ b/definitions/tools/pvacseq.cwl
@@ -21,7 +21,7 @@ arguments: [
 requirements:
     - class: ShellCommandRequirement
     - class: DockerRequirement
-      dockerPull: "susannakiwala/pvactools:4.0.0_rc_pvacview_v19" #TODO: replace with final release version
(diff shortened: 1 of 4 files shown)


definitions/tools/pvacseq.cwl
```diff
@@ -15,14 +15,14 @@ arguments: [
     { valueFrom: " && ", shellQuote: false },
     "/usr/local/bin/pvacseq", "run",
     "--iedb-install-directory", "/opt/iedb",
-    "--blastp-path", "/opt/ncbi-blast-2.12.0+/bin/blastp",
+    "--peptide-fasta", "/opt/reference_fasta/Homo_sapiens.GRCh38.pep.all.fa.gz",
```
tmooney, line 18: Would it make sense for this to be an input the user could specify but with this path as the default?

PR #1086, opened 2023-06-21
Update python and cwltool version for testing CWL syntax

tmooney:
python3.7 is about to go EOL, so let's switch to 3.10.


susannasiebert (approved):
+1

PR #1087, opened 2023-06-21
Also update the docs updater to python 3.10 and newer cwltool.

tmooney:
It would probably help to update this action, too! 🙂


susannasiebert (approved):
+1

